PR #3, opened 2026-02-03
DATA-5706 - feat(controller): add DocumentProcessor CRD for unstructured document conversion

PuneetPunamiya:
- Introduce DocumentProcessor custom resource with DoclingConfig for OCR,
      PDF backend, and table extraction settings
- Implement controller with job tracking, retry logic, and status conditions
- Add Docling client for async document conversion via docling-serve API
- Add filestore package for local caching of downloaded files


pkg/filestore/filestore.go
```diff
@@ -237,15 +254,16 @@ func (fs *FileStore) Delete(ctx context.Context, path string) error {
 		var notFound *s3types.NotFound
 		if errors.As(err, &notFound) {
 			return fmt.Errorf("file does not exist: %s", localPath)
+		} else {
+			return err
```
concaf: not needed

PR #17, opened 2026-02-18
DATA - 5711 - 	Set up E2E tests for upstream repo

gshikhar2021:
> Added e2e testing setup 

- create secrets (snowflake and aws)
- setup docling and localstack
- setup PVC for caching directory
- create controllerConfig and checked for its healthy state

>  Fixed RBAC roles when running 'make deploy' command


.github/workflows/test-e2e.yml
```diff
@@ -2,27 +2,64 @@ name: E2E Tests
...
+      - name: Create test namespace
+        run: |
+          kubectl create namespace ${{ env.NAMESPACE }}
 
-      - name: verify installation
+      - name: Verify Kind and kubectl
         run: |
           kind version
-          kubectl version
+
```
PuneetPunamiya: Any reason why kubectl version was removed ?

test/e2e/main_test.go
```diff
@@ -0,0 +1,342 @@
...
+		pf.Stdout = os.Stdout
+		pf.Stderr = os.Stderr
+		if err := pf.Start(); err != nil {
+			log.Printf("failed to port-forward docling-serve: %s", err)
+			return err
+		}
+		*runningProcesses = append(*runningProcesses, *pf)
+	}
+
+	log.Println("Applying ControllerConfig from test/e2e/config/controllerconfig.yaml...")
```
PuneetPunamiya: Since we have commented the creation of snowflake creds secret here https://github.com/redhat-data-and-ai/unstructured-data-controller/pull/17/changes#diff-345f4ae23b0645edf99172a04da688e38f03f0e398ca32ec3905d4ecc325cdccR202-R222  I think the controllerConfig won't be healthy because it will return an error here https://github.com/redhat-data-and-ai/unstructured-data-controller/blob/main/internal/controller/controllerconfig_controller.go#L133-L137

test/e2e/main_test.go
```diff
@@ -0,0 +1,342 @@
...
+		utils.RunCommand(fmt.Sprintf("kubectl get pods -n %s -l app=localstack", testNamespace))
+		log.Println("Waiting for localstack to be ready...")
+		if err := wait.For(
+			conditions.New(client.Resources()).DeploymentAvailable("localstack", testNamespace),
+			wait.WithTimeout(10*time.Minute),
+			wait.WithInterval(5*time.Second),
+		); err != nil {
+			log.Printf("Timed out waiting for localstack: %s", err)
+			log.Println("Checking localstack pod logs...")
+			utils.RunCommand(fmt.Sprintf("kubectl describe deployment localstack -n %s", testNamespace))
```
PuneetPunamiya: I don't think we need to add describe command

PR #20, opened 2026-02-24
Support auth key for docling

piyush-garg:
This is to add support for user key for
docling requests, also add fallback for
requests with both bearer key and auth key

### DIFF
--- a/api/v1alpha1/controllerconfig_types.go
+++ b/api/v1alpha1/controllerconfig_types.go
@@ -47,6 +47,7 @@ type UnstructuredDataProcessingConfigSpec struct {
 
 	IngestionBucket string `json:"ingestionBucket,omitempty"`
 	DoclingServeURL string `json:"doclingServeURL,omitempty"`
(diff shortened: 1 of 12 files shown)


api/v1alpha1/controllerconfig_types.go
```diff
@@ -47,6 +47,7 @@ type UnstructuredDataProcessingConfigSpec struct {
 
 	IngestionBucket string `json:"ingestionBucket,omitempty"`
 	DoclingServeURL string `json:"doclingServeURL,omitempty"`
+	DoclingSecret   string `json:"doclingSecret,omitempty"`
```
concaf: secret-file.env
```
DOCLING_SECRET=abc
AWS_SECRET=xyz
```

```
base64 -w0 secret-file.env > base64-secret.txt
store base64-secret.txt in github CI var
in github actions, cat base64-secret.txt > secret-file.env
kubectl create secret generic --from-env-file secret-file.env secrets
then add `envFrom: ` in deployment from this secret
```

PR #22, opened 2026-02-26
Add PVC for local caching directory

gshikhar2021:
- Added PVC for temporary caching directory to store intermediate results

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -92,7 +92,12 @@ spec:
           requests:
             cpu: 10m
             memory: 64Mi
-        volumeMounts: []
-      volumes: []
+        volumeMounts:
+        - name: unstructured-datastore-cache-storage
+          mountPath: /data/cache
+      volumes:
+      - name: unstructured-datastore-cache-storage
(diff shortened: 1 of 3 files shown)


config/manager/pvc.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: unstructured-datastore-cache
+  namespace: system
+spec:
+  accessModes:
+    - ReadWriteOnce
+  storageClassName: standard
+  resources:
+    requests:
+      storage: 10Gi
```
concaf: what's the limit in internal openshift cluster ? if possible let's increase this ... this will fill up in no time 😉

config/manager/manager.yaml
```diff
@@ -57,42 +57,47 @@ spec:
...
-        resources:
-          limits:
-            cpu: 500m
-            memory: 128Mi
-          requests:
-            cpu: 10m
-            memory: 64Mi
-        volumeMounts: []
-      volumes: []
+        - command:
```
piyush-garg: why all these changes?

PR #35, opened 2026-03-07
  Add chunking strategies using Docling

pradeepto:
Introduce two new chunking strategies — doclingHierarchicalChunker and
doclingHybridChunker — that use the Docling Serve API to chunk documents
directly from raw files via async POST /v1/chunk/{type}/source/async.

Unlike LangChain strategies which chunk pre-converted markdown, Docling's
chunk endpoint performs document conversion and chunking in a single call.
This means document processing (DocumentProcessor CR) is unnecessary when
using Docling chunking strategies.

Changes:
- Add Docling async chunking client (pkg/docling/chunk.go) — submits task, polls status, fetches result via /v1/result/{task_id}
- Add DoclingHierarchicalChunker and DoclingHybridChunker types implementing the Chunker interface
- Update Chunker interface to Chunk(ctx context.Context, input string) for context propagation
- Refactor chunkFile() and chunkRawFile() to use unified chunker.Chunk() for all strategies
- Change DocumentProcessorConfig to pointer (*DocumentProcessorConfig); skip DocumentProcessor CR creation when nil
- Add ValidateSpec() to reject DocumentProcessorConfig when using Docling chunking strategies
- Add raw file processing path in ChunksGenerator controller for Docling strategies
- Fix nil-safety in ChunksFileMetadata.Equal() for missing ConvertedFileMetadata
- Add unit tests for async chunking, Docling chunker types, spec validation, and metadata comparison

### DIFF
--- a/pkg/docling/chunk.go
+++ b/pkg/docling/chunk.go
@@ -22,7 +22,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"net/http"
 	"net/url"
 	"time"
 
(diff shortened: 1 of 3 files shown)


concaf:
closing as stale 🪰

PR #36, opened 2026-03-08
Add github action workflow for e2e tests

gshikhar2021:
- https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-secrets - Unable to access upstream repo secrets 
- workflow needs to there in the main

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -1,13 +1,22 @@
 name: E2E Tests
 
 on:
-  push:
-    branches: [ main ]
   pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+    branches: [main]
+
 jobs:
   test-e2e:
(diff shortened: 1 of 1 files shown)


.github/workflows/test-e2e.yml
```diff
@@ -1,28 +1,78 @@
...
-  pull_request:
+  pull_request_target:
+    branches: [main]
+
+env:
+  KIND_CLUSTER: unstructured-data-controller-test-e2e
+  NAMESPACE: unstructured-controller-namespace
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+  SNOWFLAKE_ACCOUNT: ${{ secrets.SNOWFLAKE_ACCOUNT }}
```
piyush-garg: why we are exposing secret as env?

PR #126, opened 2026-03-30
feat: add artifacts-based configuration for selective destination syncing

PuneetPunamiya:
* Enables selective syncing of file types with custom artifact paths under
a stages directory structure

* Users can configure which file types (converted/chunks/embeddings) to sync
and customize their destination paths

* Example usage in UnstructuredDataProduct 

```yaml
spec:
  destinationConfig:
    artifacts:
     - type: stage
       name: documentProcessorConfig
       path: processed-documents
     - type: stage
       name: chunksGeneratorConfig
       path: chunks
     - type: stage
       name: vectorEmbeddingsGeneratorConfig
       path: vector-embeddings
```


internal/controller/unstructureddataproduct_controller.go
```diff
@@ -302,6 +318,25 @@ func (r *UnstructuredDataProductReconciler) Reconcile(ctx context.Context, req c
...
+			artifactPathMap[unstructured.ChunksFileSuffix] = artifact.Path
+		case ArtifactNameVectorEmbeddings:
+			artifactPathMap[unstructured.VectorEmbeddingsFileSuffix] = artifact.Path
+		default:
+			logger.Info("unknown artifact name, skipping", "name", artifact.Name)
+		}
+	}
+	return artifactPathMap
+}
+
```
gshikhar2021: Just a suggestion , its a util function , can't we move it to controller utils??

internal/controller/unstructureddataproduct_controller.go
```diff
@@ -311,30 +346,59 @@ func (r *UnstructuredDataProductReconciler) setupSnowflakeDestination(ctx contex
...
+		Database:        unstructuredDataProductCR.Spec.DestinationConfig.SnowflakeInternalStageConfig.Database,
+		Schema:          unstructuredDataProductCR.Spec.DestinationConfig.SnowflakeInternalStageConfig.Schema,
+		ArtifactPathMap: buildArtifactPathMap(ctx, unstructuredDataProductCR.Spec.DestinationConfig.Artifacts),
 	}, nil
 }
 
 // setupS3Destination returns an S3 destination for the given CR
-func setupS3Destination(unstructuredDataProductCR *operatorv1alpha1.UnstructuredDataProduct, dataProductName string) (unstructured.Destination, error) {
+func setupS3Destination(ctx context.Context, unstructuredDataProductCR *operatorv1alpha1.UnstructuredDataProduct, dataProductName string) (unstructured.Destination, error) {
 	destCfg := unstructuredDataProductCR.Spec.DestinationConfig.S3DestinationConfig
```
gshikhar2021: can we make it controller function as we did with setupSnowflakeDestination to make it consistent and can we callable using r.func()

internal/controller/unstructureddataproduct_controller.go
```diff
@@ -311,30 +346,59 @@ func (r *UnstructuredDataProductReconciler) setupSnowflakeDestination(ctx contex
...
+		default:
+			logger.Info("unknown artifact name, skipping", "name", artifact.Name)
+		}
+
+		filesToSync = append(filesToSync, filteredFiles...)
+	}
+
+	return filesToSync
+}
+
```
gshikhar2021: Same with this , it is also a util function

pkg/unstructured/destination.go
```diff
@@ -163,12 +147,130 @@ func (d *SnowflakeInternalStage) SyncFilesToDestination(ctx context.Context,
...
+
+	if strings.HasSuffix(filePathInFilestore, ConvertedFileSuffix) {
+		suffix = ConvertedFileSuffix
+		defaultPath = DefaultProcessedDocumentsPath
+	} else if strings.HasSuffix(filePathInFilestore, ChunksFileSuffix) {
+		suffix = ChunksFileSuffix
+		defaultPath = DefaultChunksPath
+	} else if strings.HasSuffix(filePathInFilestore, VectorEmbeddingsFileSuffix) {
+		suffix = VectorEmbeddingsFileSuffix
+		defaultPath = DefaultVectorEmbeddingsPath
```
gshikhar2021: If user is defining the artifacts path then do we actually need to use fallback mechanism for constructing paths , because at later stage we checking the existence of that path if it mismatches then ?

pkg/unstructured/destination.go
```diff
@@ -178,26 +280,34 @@ func (d *S3Destination) getPrefix() string {
...
+	// Determine artifact path based on file suffix
+	artifactPath := getArtifactPathForFile(filePathInFilestore, d.ArtifactPathMap)
+
+	// Build path: prefix/stages/{artifact-path}/{filename}
+	var key string
 	if prefix != "" {
-		key = filepath.Join(prefix, baseName)
+		key = filepath.Join(prefix, "stages", artifactPath, baseName)
+	} else {
+		key = filepath.Join("stages", artifactPath, baseName)
```
gshikhar2021: Same with this , why do we need else case

internal/controller/documentprocessor_controller.go
```diff
@@ -186,6 +189,26 @@ func (r *DocumentProcessorReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+				return err
+			}
+			return controllerutils.AddForceReconcileLabel(ctx, r.Client, unstructuredDataProductCR)
+		}); err != nil {
+			logger.Error(err, "failed to add force reconcile label to UnstructuredDataProduct CR")
+			return r.handleError(ctx, documentProcessorCR, err)
+		}
+		logger.Info("triggered UnstructuredDataProduct to sync converted files")
+	}
+
```
gshikhar2021: Not required as we are using force reconcile from vector embeddings that will automatically take care

internal/controller/chunksgenerator_controller.go
```diff
@@ -173,6 +174,26 @@ func (r *ChunksGeneratorReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+				return err
+			}
+			return controllerutils.AddForceReconcileLabel(ctx, r.Client, unstructuredDataProductCR)
+		}); err != nil {
+			logger.Error(err, "failed to add force reconcile label to UnstructuredDataProduct CR")
+			return r.handleError(ctx, chunksGeneratorCR, err)
+		}
+		logger.Info("triggered UnstructuredDataProduct to sync chunks files")
+	}
+
```
gshikhar2021: This is also not required , taken care by vector embeddings controller

- refer this branch https://github.com/gshikhar2021/unstructured-data-controller/tree/artifacts

internal/controller/unstructureddataproduct_controller.go
```diff
@@ -227,20 +230,32 @@ func (r *UnstructuredDataProductReconciler) Reconcile(ctx context.Context, req c
...
+				return err
+			}
+			return controllerutils.AddForceReconcileLabel(ctx, r.Client, latestDocumentProcessorCR)
+		}); err != nil {
+			logger.Error(err, "failed to add force reconcile label to DocumentProcessor CR")
+			return r.handleError(ctx, unstructuredDataProductCR, err)
+		}
+		logger.Info("triggered DocumentProcessor for unprocessed files")
+	} else {
+		logger.Info("no unprocessed files, skipping DocumentProcessor trigger")
```
gshikhar2021: Revert back to the previous logic as document processor will automatically skip all the processed files

internal/controller/unstructureddataproduct_controller.go
```diff
@@ -344,6 +408,27 @@ func (r *UnstructuredDataProductReconciler) SetupWithManager(mgr ctrl.Manager) e
...
+
+	// Check if any raw file doesn't have a converted file
+	for _, rawFile := range rawFiles {
+		if !convertedMap[rawFile] {
+			return true
+		}
+	}
+	return false
+}
+
```
gshikhar2021: Remove this function as it is no longer required

pkg/unstructured/destination.go
PuneetPunamiya: The fallback mechanism is needed for user convenience when they want to sync artifacts but don't specify a custom path.
                                                                                                                                                                                                                    
  **Use cases:**
                                                                                                                                                                                                                    
  1. **User specifies custom path:**
  ```yaml
  artifacts:                                                                                                                                                                                                        
    - type: stage
      name: chunksGeneratorConfig                                                                                                                                                                                   
      path: "my-custom-chunks"
```
  → Uses custom path: stages/my-custom-chunks/file.json                                                                                                                                                             
                                                                                                                                                                                                                    
  2. User omits path or provides empty string:    
  ```yaml                                                                                                                                                                  
  artifacts:                                                                                                                                                                                                        
    - type: stage 
      name: chunksGeneratorConfig
```
  → Falls back to default: stages/chunks/file.json                                                                                                                                                                  
  
  This allows users to:                                                                                                                                                                                             
  - Declare what to sync (via the artifacts list)
  - Optionally customize where to sync (via the path field)                                                                                                                                                         
  - Get sensible defaults when path is not specified       
                                                                                                                                                                                                                    
  Could you clarify what you mean by "checking the existence of that path if it mismatches"? I don't see any path validation logic in the sync flow - we only check if files already exist at the destination to    
  avoid redundant uploads.

internal/controller/documentprocessor_controller.go
PuneetPunamiya: Good catch !

internal/controller/unstructureddataproduct_controller.go
PuneetPunamiya: Since it is controller based and won't be reused again hence did not add in the controller utils

internal/controller/unstructureddataproduct_controller.go
PuneetPunamiya: same as above

api/v1alpha1/unstructureddatapipeline_types.go
```diff
@@ -115,9 +115,17 @@ type S3Config struct {
...
+type ArtifactConfig struct {
+	Type string `json:"type,omitempty"` // e.g., "stage"
+	Name string `json:"name,omitempty"` // e.g., "documentProcessorConfig", "chunksGeneratorConfig", "vectorEmbeddingsGeneratorConfig"
+	Path string `json:"path,omitempty"` // e.g., "processed-documents", "chunks", "vector-embeddings"
+}
+
 // DestinationConfig defines where to write processed data (e.g. Snowflake internal stage or S3).
 type DestinationConfig struct {
 	Type                         UnstructuredDataType         `json:"type,omitempty"`
+	Artifacts                    []ArtifactConfig             `json:"artifacts,omitempty"`
```
piyush-garg: should we make it required with at least one document to upload

internal/controller/unstructureddataproduct_controller.go
```diff
@@ -302,6 +311,25 @@ func (r *UnstructuredDataProductReconciler) Reconcile(ctx context.Context, req c
 	}, nil
 }
 
+// buildArtifactPathMap builds a map from file suffix to artifact path based on the artifacts configuration
+func buildArtifactPathMap(ctx context.Context, artifacts []operatorv1alpha1.ArtifactConfig) map[string]string {
```
piyush-garg: should we do the naming convention like `stage name` is same as `path` then we dont need to do this, whatever the stages or paths mentioned in artifacts, we iterate over those path and upload

pkg/unstructured/destination.go
```diff
@@ -163,12 +147,130 @@ func (d *SnowflakeInternalStage) SyncFilesToDestination(ctx context.Context,
...
+	}
+
+	if path, ok := artifactPathMap[suffix]; ok && path != "" {
+		return path
+	}
+	return defaultPath
+}
+
+// filesAreEqual compares files based on their type and returns true if metadata is unchanged
+func filesAreEqual(filePath string, existingData string, newData []byte) bool {
```
piyush-garg: why this is not generic

gshikhar2021:
@PuneetPunamiya , can we remove Equal Function from converted_file.go, chunks_file.go, embeddings_file.go because we no longer required that ?


- Also in samples file , can we provide sample config because from the new user's POV , it is difficult to write specs, may be you comment it out

PuneetPunamiya:
> @PuneetPunamiya , can we remove Equal Function from converted_file.go, chunks_file.go, embeddings_file.go because we no longer required that ?
> 
>     * Also in samples file , can we provide sample config because from the new user's POV , it is difficult to write specs, may be you comment it out


@gshikhar2021 those functions are used in the respective controllers code

Regarding samples file, will update the samples

internal/controller/artifact_helper.go
```diff
@@ -0,0 +1,17 @@
+package controller
+
+import "github.com/redhat-data-and-ai/unstructured-data-controller/pkg/unstructured"
+
+// GetProcessorForArtifact returns the artifact processor for the given artifact name
+func GetProcessorForArtifact(artifactName string) unstructured.ArtifactProcessor {
```
piyush-garg, line 6: why we need this ?

pkg/unstructured/file.go
```diff
@@ -33,6 +33,11 @@ type RawFileMetadata struct {
 	UID      string `json:"uid,omitempty"`
 }
 
+type ArtifactProcessor interface {
+	GetFileSuffix() string
+	CompareMetadata(existingData string, newData []byte) bool
```
piyush-garg: why this needs to be written for every controller ?

PuneetPunamiya: updated!

internal/controller/unstructureddatapipeline_controller.go
```diff
@@ -242,13 +245,14 @@ func (r *UnstructuredDataPipelineReconciler) Reconcile(ctx context.Context, req
 		logger.Error(err, "failed to add force reconcile label to DocumentProcessor CR")
 		return r.handleError(ctx, unstructuredDataPipelineCR, err)
 	}
+	logger.Info("triggered DocumentProcessor for unprocessed files")
```
gshikhar2021: Adding force reconcile for document processor anyways , logger statement is misleading, either remove this or change it to "force reconcile label successfully added" or something like that

PuneetPunamiya: updated!

api/v1alpha1/unstructureddatapipeline_types.go
gshikhar2021: @PuneetPunamiya , this is still needs to be taken care

internal/controller/artifact_helper.go
```diff
@@ -0,0 +1,30 @@
...
+	case "documentProcessorConfig":
+		return unstructured.DefaultProcessedDocumentsPath
+	case "chunksGeneratorConfig":
+		return unstructured.DefaultChunksPath
+	case "vectorEmbeddingsGeneratorConfig":
+		return unstructured.DefaultVectorEmbeddingsPath
+	default:
+		return ""
+	}
+}
```
gshikhar2021: I think this is making the code complex, can't we do like this
```
type ArtifactNameType string

const (
	DocumentProcessorConfig         ArtifactNameType = "documentProcessorConfig"
	ChunksGeneratorConfig           ArtifactNameType = "chunksGeneratorConfig"
	VectorEmbeddingsGeneratorConfig ArtifactNameType = "vectorEmbeddingsGeneratorConfig"
)

func (a ArtifactNameType) GetSuffix() string {
	switch a {
	case DocumentProcessorConfig:
		return ConvertedFileSuffix
	case ChunksGeneratorConfig:
		return ChunksFileSuffix
	case VectorEmbeddingsGeneratorConfig:
		return VectorEmbeddingsFileSuffix
	default:
		return ""
	}
}

func (a ArtifactNameType) GetDefaultPath() string {
	switch a {
	case DocumentProcessorConfig:
		return DefaultProcessedDocumentsPath
	case ChunksGeneratorConfig:
		return DefaultChunksPath
	case VectorEmbeddingsGeneratorConfig:
		return DefaultVectorEmbeddingsPath
	default:
		return ""
	}
}

func (a ArtifactNameType) IsValid() bool {
	return a.GetSuffix() != ""
}
```

so eventually all the unneccessary interfaces inside the artifact_helper.go , DocumentProcessor , chunks generator  will be eliminated
```
type ChunksGenerator struct{}

func (ChunksGenerator) GetFileSuffix() string {
	return unstructured.ChunksFileSuffix
}
```
wdyt ???

pkg/unstructured/destination.go
```diff
@@ -20,139 +22,137 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+const (
```
piyush-garg: why these constants are defined here ?

pkg/unstructured/destination.go
```diff
@@ -20,139 +22,137 @@ import (
...
-	Role     string
-	Database string
-	Schema   string
-	Stage    string
+	Client          *snowflake.Client
+	Role            string
+	Database        string
+	Schema          string
+	Stage           string
+	DataProductName string
```
piyush-garg: why we need this?

api/v1alpha1/unstructureddatapipeline_types.go
PuneetPunamiya: updated!

internal/controller/unstructureddatapipeline_controller.go
```diff
@@ -335,6 +342,42 @@ func setupS3Destination(unstructuredDataPipelineCR *operatorv1alpha1.Unstructure
 	}, nil
 }
 
+func buildDestinationPath(ctx context.Context, unstructuredDataProductCR operatorv1alpha1.UnstructuredDataPipeline, filePaths []string) []unstructured.ArtifactFiles {
+	logger := log.FromContext(ctx)
+	var artifactFileGroups []unstructured.ArtifactFiles
```
piyush-garg, line 347: what is artifactFileGroup? is it not containing list of files ?

internal/controller/unstructureddatapipeline_controller.go
```diff
@@ -268,22 +268,27 @@ func (r *UnstructuredDataPipelineReconciler) Reconcile(ctx context.Context, req
 		return r.handleError(ctx, unstructuredDataPipelineCR, err)
 	}
 
-	// list all files in the filestore for the data product
+	// list files to check if processing is needed
 	filePaths, err := r.fileStore.ListFilesInPath(ctx, dataProductName)
```
piyush-garg, line 272: why it is done here ?

internal/controller/unstructureddatapipeline_controller.go
```diff
@@ -335,6 +342,42 @@ func setupS3Destination(unstructuredDataPipelineCR *operatorv1alpha1.Unstructure
...
+
+		var artifactFiles []string
+		for _, filePath := range filePaths {
+			if strings.HasSuffix(filePath, suffix) {
+				artifactFiles = append(artifactFiles, filePath)
+			}
+		}
+
+		artifactPath := artifact.Path
+		if artifactPath == "" {
```
piyush-garg: why can we check suffix and path together in single loop ?

pkg/unstructured/destination.go
```diff
@@ -167,12 +159,87 @@ func (d *SnowflakeInternalStage) SyncFilesToDestination(ctx context.Context,
...
+	}
+
+	if filepath.Separator != '/' {
+		path = filepath.ToSlash(path)
+	}
+	return path
+}
+
+// filesAreEqual compares files using generic JSON comparison
+func filesAreEqual(existingData string, newData []byte) bool {
```
piyush-garg: should we compare using some other identifier etc ?

internal/controller/unstructureddatapipeline_controller.go
PuneetPunamiya: Here ArtifactFileGroup is one configured artifact i.e. 
* Files - list of file paths in filestore
* Path - destination path 

Shall I rename it to something like `ArtifactSyncGroup`

pkg/unstructured/destination.go
```diff
@@ -167,12 +157,116 @@ func (d *SnowflakeInternalStage) SyncFilesToDestination(ctx context.Context,
...
+		path = filepath.Join("stages", artifactPath, baseName)
+	} else {
+		path = baseName
+	}
+
+	if filepath.Separator != '/' {
+		path = filepath.ToSlash(path)
+	}
+	return path
+}
```
gshikhar2021, line 217: here we are creating global buildDestinationPath Method but for s3 we are defining same method on struct type creating inconsistency, can we do this like

- func (d *SnowflakeInternalStage) buildDestinationPath(filePathInFilestore string, artifactPath string) string{}

pkg/unstructured/destination.go
```diff
@@ -167,12 +157,116 @@ func (d *SnowflakeInternalStage) SyncFilesToDestination(ctx context.Context,
...
+
+	// Build path: {extractedPrefix}/stages/{artifact-path}/{filename}
+	var path string
+	if extractedPrefix != "" && artifactPath != "" {
+		path = filepath.Join(extractedPrefix, "stages", artifactPath, baseName)
+	} else if artifactPath != "" {
+		path = filepath.Join("stages", artifactPath, baseName)
+	} else {
+		path = baseName
+	}
```
gshikhar2021, line 211: also here we are constructing path using 

- DataProduct/prefix/stages/path/file_name
but in s3 sync we are omiting this prefix and keeping like this 
- DataProduct/stages/path/file_name

is this intentional ?

test/e2e/unstructured_test.go
```diff
@@ -738,6 +767,348 @@ func TestUnstructuredDataLoad(t *testing.T) {
...
+
+					for suffix, expectedPathSegment := range expectedPathBySuffix {
+						if !strings.HasSuffix(normalizedPath, suffix) {
+							continue
+						}
+						if !strings.Contains(normalizedPath, expectedPathSegment) {
+							t.Logf("artifact found in unexpected path: %s (expected to contain: %s)", normalizedPath, expectedPathSegment)
+							return false, nil
+						}
+						matchedExpectedSuffixes[suffix] = true
```
gshikhar2021, line 830: Test expects -  /stages/chunks/

Using strings.Contains() matches substring
- unstructured/source-data/stages/chunks/file.json will pass
- random/stages/chunks/file.json - wrong prefix, but passes

just replace the above line 779 and other occurences of it to the full path

pkg/unstructured/destination.go
```diff
@@ -167,12 +157,116 @@ func (d *SnowflakeInternalStage) SyncFilesToDestination(ctx context.Context,
...
+func buildDestinationPath(filePathInFilestore string, artifactPath string) string {
+	baseName := filepath.Base(filePathInFilestore)
+
+	// filePathInFilestore example: "testunstructureddataproduct/source-data/file.pdf-converted.json"
+	// "testunstructureddataproduct/source-data"
+	var extractedPrefix string
+	if idx := strings.LastIndex(filePathInFilestore, "/"); idx != -1 {
+		extractedPrefix = filePathInFilestore[:idx]
+	}
+
```
gshikhar2021, line 202: here we are extracting out the prefix using last '/'index , what if there is '/' in the file name ??????
can't we take dp name and prefix from CR and create the complete prefix???

pkg/unstructured/destination.go
PuneetPunamiya: Yes because in snowflake destination we do not have a prefix field

PR #137, opened 2026-04-08
Add make targets for local development setup and documentation

PuneetPunamiya:
* Add  DEVELOPMENT.md guide covering the complete local development
workflow including LocalStack setup, CRD installation, controller
configuration, and e2e testing with required environment variables

* Introduce make targets (local-dev-setup, local-dev-cleanup) with
corresponding shell scripts to automate local environment creation,
including S3 buckets, SQS queues, and Docling service setup.

* Fix test-e2e target to pass KIND_CLUSTER and SKIP_CLUSTER_SETUP
environment variables, preventing duplicate cluster creation and
aligning local test execution with CI workflow.


DEVELOPMENT.md
```diff
@@ -0,0 +1,192 @@
...
+
+Start LocalStack in a separate terminal to simulate AWS services locally:
+
+```bash
+sudo docker run --rm -it \
+  -p 127.0.0.1:4566:4566 \
+  -p 127.0.0.1:4510-4559:4510-4559 \
+  docker.io/localstack/localstack
+```
+
```
gshikhar2021: just fyi , localstack latest image requires auth , its better to use specific release
sudo docker run --rm -it \
  -p 127.0.0.1:4566:4566 \
  -p 127.0.0.1:4510-4559:4510-4559 \
  docker.io/localstack/localstack:4.14.0

scripts/local-dev-setup.sh
```diff
@@ -0,0 +1,103 @@
...
+else
+    awslocal sqs create-queue --queue-name "${SQS_QUEUE_NAME}"
+    echo "✓ SQS queue '${SQS_QUEUE_NAME}' created"
+fi
+
+# Configure S3 bucket notification
+echo ""
+echo "5c. Configuring S3 bucket notification..."
+awslocal s3api put-bucket-notification-configuration \
+    --bucket "${DATA_INGESTION_BUCKET}" \
```
gshikhar2021: Instead of directly applying notification.json file there might be the possibility that there is a queueUrl mismatch like in mac , queue url is "QueueUrl": "http://sqs.us-west-2.localhost.localstack.cloud:4566/000000000000/unstructured-s3-queue" and for other user it might be us-east-1
So in my opinion there are 2 solutions either 
- Read the url from last step and then put notification
- Add additional step to explicitly tell use that they should use specific region like us-west-2 or us-east-1 and while creating the queue mention the region explicitly

DEVELOPMENT.md
```diff
@@ -0,0 +1,192 @@
...
+## Step 3: Create the Unstructured Secret
+
+The controller requires AWS credentials, Snowflake private key, and Docling authentication stored in a Kubernetes secret.
+
+### Apply the Secret
+
+A template exists at `test/resources/unstructured/unstructured-secret.yaml`. Apply it with:
+
+```bash
```
gshikhar2021, line 48: In my opinion there are multiple secrets in unstructured_secret.yaml file 
- can we tell user to use localhost instead of localstack in SOURCE_AWS_ENDPOINT: http://localstack:4566 and other AWS endpoints , wdyt????

DEVELOPMENT.md
```diff
@@ -0,0 +1,192 @@
...
+    app.kubernetes.io/managed-by: kustomize
+  name: sqsinformer-sample
+spec:
+  queueURL: http://sqs.us-east-1.localhost.localstack.cloud:4566/000000000000/unstructured-s3-queue
+```
+
+### Apply the SQSInformer
+
+```bash
+kubectl apply -f config/samples/operator_v1alpha1_sqsinformer.yaml
```
gshikhar2021: use namespace as well 

-  -n unstructured-controller-namespace

DEVELOPMENT.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Verification
+
+Monitor controller logs for reconciliation events and verify that resources are ready:
+
+### Check ControllerConfig Status
+
+```bash
+kubectl get controllerconfig controllerconfig-sample -o yaml
+```
```
gshikhar2021: what about adding namespace as well

DEVELOPMENT.md
```diff
@@ -0,0 +1,192 @@
...
+
+Should display "ConfigReady" condition as true.
+
+### Check SQSInformer Status
+
+```bash
+kubectl get sqsinformer sqsinformer-sample -o yaml
```
gshikhar2021: same here namespace ?

scripts/local-dev-setup.sh
PuneetPunamiya: Have updated the PR

DEVELOPMENT.md
PuneetPunamiya: Done!

DEVELOPMENT.md
```diff
@@ -0,0 +1,200 @@
+# Local Development Setup
+
+This guide walks you through setting up the Unstructured Data Controller for local development.
+
+## Prerequisites
+
+- Docker
+- kubectl
+- Go
+
```
gshikhar2021, line 15: just add docling-serve also in pre-requisites

PuneetPunamiya: We are running docling as a container right ? And it will be running in background as well right ?

gshikhar2021: Just wanted to check where are we doing that in that the setup script , could you please point it out ?

PuneetPunamiya: I think you are right, will update the PR

scripts/local-dev-cleanup.sh
```diff
@@ -0,0 +1,32 @@
...
+    CACHE_DIR="${REPO_ROOT}/${CACHE_DIR}"
+fi
+
+# Configuration
+LOCAL_KIND_CLUSTER="${LOCAL_KIND_CLUSTER:-unstructured-data-controller-local}"
+
+echo "Cleaning up local development environment..."
+
+# Stop Docling if it was started by setup script
+if [ -f /tmp/docling.pid ]; then
```
piyush-garg, line 11: we are stopping docling, but what about localstack and ollama

scripts/local-dev-setup.sh
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/env bash
+set -e
+
+export AWS_DEFAULT_REGION="${AWS_DEFAULT_REGION:-us-east-1}"
```
piyush-garg: can we move this to yaml

scripts/local-dev-setup.sh
```diff
@@ -0,0 +1,113 @@
...
+
+# Configuration
+LOCAL_KIND_CLUSTER="${LOCAL_KIND_CLUSTER:-unstructured-data-controller-local}"
+LOCAL_NAMESPACE="${LOCAL_NAMESPACE:-unstructured-controller-namespace}"
+CACHE_DIR="${CACHE_DIR:-tmp/cache}"
+DATA_STORAGE_BUCKET="${DATA_STORAGE_BUCKET:-data-storage-bucket}"
+DATA_INGESTION_BUCKET="${DATA_INGESTION_BUCKET:-data-ingestion-bucket}"
+OUTPUT_RESULT_BUCKET="${OUTPUT_RESULT_BUCKET:-output-result-bucket}"
```
piyush-garg: why such name?

can we follow the convention defined

piyush-garg: why we are introducing these variables in script, these should be part of yaml

this is defining them again, we can keep kind and namespace thing

scripts/local-dev-setup.sh
```diff
@@ -0,0 +1,113 @@
...
+
+# 4. Check/Start Docling
+echo ""
+echo "4. Checking Docling service..."
+if command -v docling-serve &> /dev/null; then
+    if curl -s http://localhost:5001/health &>/dev/null; then
+        echo "✓ Docling is already running"
+    else
+        echo "Starting Docling in the background..."
+        nohup docling-serve run --enable-ui > /tmp/docling.log 2>&1 &
```
piyush-garg, line 58: this will create dep to install this and not sure how it will work on other system, can we do via sh

scripts/local-dev-setup.sh
```diff
@@ -0,0 +1,113 @@
...
+    fi
+else
+    echo "⚠ Docling is not installed. You'll need to start it manually."
+fi
+
+# 5. Setting up LocalStack resources
+# Create S3 buckets
+echo ""
+echo "5a. Creating S3 buckets..."
+for bucket in "${DATA_STORAGE_BUCKET}" "${DATA_INGESTION_BUCKET}"; do
```
piyush-garg: what about file store bucket

scripts/local-dev-setup.sh
```diff
@@ -0,0 +1,113 @@
...
+fi
+
+# Configure S3 bucket notification
+echo ""
+echo "5c. Configuring S3 bucket notification..."
+QUEUE_URL="$(awslocal sqs get-queue-url --queue-name "${SQS_QUEUE_NAME}" --query 'QueueUrl' --output text)"
+QUEUE_ARN="$(awslocal sqs get-queue-attributes --queue-url "${QUEUE_URL}" --attribute-names QueueArn --query 'Attributes.QueueArn' --output text)"
+NOTIFICATION_FILE="$(mktemp)"
+trap "rm -f \"${NOTIFICATION_FILE}\"" EXIT
+cat >"${NOTIFICATION_FILE}" <<EOF
```
piyush-garg: why we are not using the file already existing

scripts/local-dev-setup.sh
```diff
@@ -0,0 +1,113 @@
...
+# Create SQS queue
+echo ""
+echo "5b. Creating SQS queue..."
+if awslocal sqs get-queue-url --queue-name "${SQS_QUEUE_NAME}" &>/dev/null; then
+    echo "✓ SQS queue '${SQS_QUEUE_NAME}' already exists"
+else
+    awslocal sqs create-queue --queue-name "${SQS_QUEUE_NAME}"
+    echo "✓ SQS queue '${SQS_QUEUE_NAME}' created"
+fi
+
```
piyush-garg: dont we need to start localstack before this

scripts/local-dev-setup.sh
```diff
@@ -0,0 +1,113 @@
...
+      "Events": ["s3:ObjectCreated:*", "s3:ObjectRemoved:*"]
+    }
+  ]
+}
+EOF
+awslocal s3api put-bucket-notification-configuration \
+    --bucket "${DATA_INGESTION_BUCKET}" \
+    --notification-configuration "file://${NOTIFICATION_FILE}"
+echo "✓ S3 bucket notification configured (queue ARN from LocalStack)"
+
```
piyush-garg, line 159: what about ollama setup

scripts/local-dev-setup.sh
```diff
@@ -0,0 +1,113 @@
...
+echo ""
+echo "✓ Local development environment setup complete!"
```
piyush-garg: dont we need to create create and controller config also, or we can do till make install and run

DEVELOPMENT.md
```diff
@@ -0,0 +1,201 @@
...
+```
+
+## Step 5: Run the Controller
+
+Launch the controller locally for development:
+
+```bash
+make run
+```
+
```
piyush-garg, line 63: make local dev should cover till here i think, i will let @concaf chime in

DEVELOPMENT.md
```diff
@@ -0,0 +1,201 @@
...
+## Step 7: Create SQSInformer Custom Resource
+
+This resource enables the controller to listen to message queue events. The queue URL must align with your LocalStack setup.
+
+### Sample SQSInformer
+
+```yaml
+apiVersion: operator.dataverse.redhat.com/v1alpha1
+kind: SQSInformer
```
piyush-garg: this can be created also, as queue is getting created in script

DEVELOPMENT.md
```diff
@@ -0,0 +1,201 @@
...
+   - Verify data is being ingested into Snowflake
+
+## Running Tests
+
+To run the end-to-end tests, you need to set up the required environment variables and build/push the Docker image.
+
+### Export Environment Variables
+
+```bash
+  export IMG=<your-image-registry>/<image-name>:<tag>
```
piyush-garg: we can take these variables as input and create controller config and secret also

piyush-garg (changes requested):

scripts/local-dev-cleanup.sh
PuneetPunamiya: updated!

.github/workflows/release.yml
concaf, line 1: not needed

hack/release.sh
concaf, line 1: same

scripts/localstack.sh
concaf, line 1: kill localstack, move to the real deal 🙈

scripts/local-dev-cleanup.sh
concaf, line 1: use kube deployments for docling-serve and embeddings

PR #141, opened 2026-04-10
e2e test enhancement

gshikhar2021:
- removed using grep command on logs
- Using exec command with timeout

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -24,6 +24,7 @@ import (
 	"log"
 	"os"
 	"os/exec"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


test/e2e/main_test.go
```diff
@@ -239,22 +240,37 @@ func testSetup(_ context.Context, runningProcesses *[]exec.Cmd, config *envconf.
...
-		setupCmd := fmt.Sprintf(`kubectl exec -n %s %s -- sh -c 'ollama pull nomic-embed-text && until ollama list | grep -q nomic-embed-text; do echo "Waiting for model..."; sleep 2; done && ollama cp nomic-embed-text nomic-ai/nomic-embed-text-v1.5'`, testNamespace, ollamaPod)
-		if p := utils.RunCommand(setupCmd); p.Err() != nil {
-			log.Printf("Failed to setup model: %s", p.Err())
-			return p.Err()
+		// Pull model with 10-minute timeout
+		log.Println("Pulling nomic-embed-text model (this may take several minutes)...")
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
+		defer cancel()
+
+		pullCmd := exec.CommandContext(ctx, "kubectl", "exec", "-n", testNamespace, podName, "--", "ollama", "pull", "nomic-embed-text:latest")
```
piyush-garg, line 259: whats the difference between previous implementation and this implementation

gshikhar2021: I have checked that we dont need separate readiness probe in the code because pod will be in the ready state only when all deployment's readiness probe becomes true

test/e2e/main_test.go
```diff
@@ -239,22 +240,37 @@ func testSetup(_ context.Context, runningProcesses *[]exec.Cmd, config *envconf.
...
+		if output, err := pullCmd.CombinedOutput(); err != nil {
+			return fmt.Errorf("failed to pull model: %v, output: %s", err, string(output))
+		}
+		log.Println("Model pulled successfully")
+
+		// Copy model with 10-minute timeout
+		ctx2, cancel2 := context.WithTimeout(context.Background(), 10*time.Minute)
```
piyush-garg: should be less, cp should not take 10 minutes, and should we need the context things, we and just hit the command

gshikhar2021: Yes you are right cp command is instantaneous, so we dont need ctx at all

piyush-garg (approved):

PR #148, opened 2026-04-21
Fix/snowflake file upload

gshikhar2021:
- It allows limited special characters alphanumeric, forward slashes, hyphens, underscores, and periods
 - replacing all other special characters with underscore(_) and converting (__) to _


piyush-garg:
Lets brainstorm this, we need way to make the file name unique  in nested struct, probably some fileId variable etc

PR #149, opened 2026-04-22
fix: Same file name for different files does not process files

gshikhar2021:
- Issue was when uploading different files with same name then it was not processing
- Previously comparing filepath in metadata to skip same content files
- Now added etag which will make sure if different files with same name is uploaded then it will check by considering etag as well

### DIFF
--- a/api/v1alpha1/documentprocessor_types.go
+++ b/api/v1alpha1/documentprocessor_types.go
@@ -42,6 +42,7 @@ type DocumentProcessorStatus struct {
 
 type Job struct {
 	FilePath          string        `json:"filePath,omitempty"`
+	FileIdentifier    string        `json:"fileIdentifier,omitempty"`
 	DocumentConverter string        `json:"documentConverter,omitempty"`
 	DoclingConfig     DoclingConfig `json:"doclingConfig,omitempty"`
 	TaskID            string        `json:"taskID,omitempty"`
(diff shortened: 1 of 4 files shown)


internal/controller/documentprocessor_controller.go
```diff
@@ -363,6 +377,7 @@ func (r *DocumentProcessorReconciler) processDocument(ctx context.Context, rawFi
 		}
 		latest.AddOrUpdateJob(operatorv1alpha1.Job{
 			FilePath:          rawFilePath,
+			FileETag:          metadataFile.UID,
```
piyush-garg: hey i am not sure how UID will help, that should remain same for the filechanges

gshikhar2021: As we have tested etag is getting changed, so it will

api/v1alpha1/documentprocessor_types.go
```diff
@@ -42,6 +42,7 @@ type DocumentProcessorStatus struct {
 
 type Job struct {
 	FilePath          string        `json:"filePath,omitempty"`
+	JobIdentifier     string        `json:"jobIdentifier,omitempty"`
```
PuneetPunamiya: Can you fix this field name

internal/controller/documentprocessor_controller.go
```diff
@@ -330,6 +331,19 @@ func (r *DocumentProcessorReconciler) processDocument(ctx context.Context, rawFi
 		return nil
 	}
 
+	// read the metadata file to get the file's UID (ETag)
+	metadataFileRaw, err := r.fileStore.Retrieve(ctx, unstructured.GetMetadataFilePath(rawFilePath))
```
piyush-garg: better to export this as function

piyush-garg: it can be reused at two places, we can mention as getFileUID

gshikhar2021: exported as the controller function

piyush-garg (approved):

PR #150, opened 2026-04-22
fix(s3): avoid re-uploading unchanged files to destination

PuneetPunamiya:
* Add checksum comparison to detect when files haven't changed. This prevents unnecessary uploads every reconciliation cycle.

### DIFF
--- a/pkg/unstructured/destination.go
+++ b/pkg/unstructured/destination.go
@@ -3,13 +3,16 @@ package unstructured
 import (
 	"bytes"
 	"context"
-	"crypto/md5"
+	"crypto/sha256"
+	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"path/filepath"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
+	"github.com/aws/aws-sdk-go-v2/service/s3/types"
 	"github.com/redhat-data-and-ai/unstructured-data-controller/pkg/awsclienthandler"
 	"github.com/redhat-data-and-ai/unstructured-data-controller/pkg/filestore"
 	"github.com/redhat-data-and-ai/unstructured-data-controller/pkg/snowflake"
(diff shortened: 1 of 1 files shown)


pkg/unstructured/destination.go
```diff
@@ -236,26 +238,31 @@ func (d *S3Destination) SyncFilesToDestination(ctx context.Context, fs *filestor
...
-			Key:    aws.String(key),
+			Bucket:       aws.String(d.Bucket),
+			Key:          aws.String(key),
+			ChecksumMode: types.ChecksumModeEnabled,
 		})
-		if err == nil && headResp.ETag != nil {
-			// S3 ETag is MD5 hash wrapped in quotes for single-part uploads
-			localMD5 := fmt.Sprintf("\"%x\"", md5.Sum(data))
-			if localMD5 == *headResp.ETag {
+
```
piyush-garg, line 252: do we need to handle the scenario of err here

pkg/unstructured/destination.go
```diff
@@ -236,26 +238,31 @@ func (d *S3Destination) SyncFilesToDestination(ctx context.Context, fs *filestor
...
-		// File is new or changed, upload it
+		// File is new or changed, upload it with SHA256
 		_, err = s3Client.PutObject(ctx, &s3.PutObjectInput{
-			Bucket: aws.String(d.Bucket),
-			Key:    aws.String(key),
-			Body:   bytes.NewReader(data),
+			Bucket:            aws.String(d.Bucket),
+			Key:               aws.String(key),
+			Body:              bytes.NewReader(data),
+			ChecksumAlgorithm: types.ChecksumAlgorithmSha256,
```
piyush-garg, line 273: if we can add test for this, that will be good

pkg/unstructured/destination.go
PuneetPunamiya: Yeah we can handle the error here 

```golang
if err != nil {
	var notFound *types.NotFound
	if !errors.As(err, &notFound) {
		logger.Info("unable to check existing file, will upload anyway",
			"key", key, "error", err.Error())
	}
} else if headResp.ChecksumSHA256 != nil {
	if *headResp.ChecksumSHA256 == localSHA256 {
		logger.Info("file unchanged, skipping upload",
			"file", chunksFilePath, "key", key)
		continue
	}
}
```

I'm thinking to handle the error in this way

piyush-garg: lets do as we discussed

piyush-garg (approved):

PR #160, opened 2026-04-24
Make pvc storage size configurable dynamically

gshikhar2021:


config/manager/pvc.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
   name: datastore-cache
+  namespace: system
```
piyush-garg: this is not required

config/deploy/kustomization.yaml
```diff
@@ -12,5 +13,19 @@ images:
 - name: controller
   newName: ghcr.io/redhat-data-and-ai/unstructured-data-controller
   newTag: latest
+patches:
```
piyush-garg: if we are exporting as env, then we dont need to add patch

piyush-garg (changes requested):

config/manager/pvc.yaml
gshikhar2021: removed

config/deploy/kustomization.yaml
gshikhar2021: removed patch

Makefile
```diff
@@ -218,12 +219,12 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
 .PHONY: deploy
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/deploy && $(KUSTOMIZE) edit set image controller=${IMG} && $(KUSTOMIZE) edit set namespace "${DEPLOYMENT_NAMESPACE}" && $(KUSTOMIZE) edit set namesuffix -- "${SUFFIX}"
-	$(KUSTOMIZE) build config/deploy | $(KUBECTL) apply -f -
+	export PVC_SIZE=${PVC_SIZE} && $(KUSTOMIZE) build config/deploy | envsubst | $(KUBECTL) apply -f -
```
piyush-garg: should we run evsubst only on pvc.yaml instead the complete release yaml

config/deploy/kustomization.yaml
```diff
@@ -4,6 +4,7 @@ namePrefix: unstructured-
 
 namespace: unstructured-controller-namespace
```
piyush-garg, line 6: this file changes can be cleaned

gshikhar2021: cleaned

Makefile
gshikhar2021: there are 2 approaches to do this 
- use in-place substitution using sponge, but afterwards need to restore the file because of the substituted value (Used this in my code)
- use temp file substitution  by copying pvc.yaml to temp file and then back to pvc.yaml

Makefile
```diff
@@ -217,13 +218,17 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
...
 .PHONY: deploy
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
+	PVC_SIZE=${PVC_SIZE} envsubst < config/manager/pvc.yaml | sponge config/manager/pvc.yaml
 	cd config/deploy && $(KUSTOMIZE) edit set image controller=${IMG} && $(KUSTOMIZE) edit set namespace "${DEPLOYMENT_NAMESPACE}" && $(KUSTOMIZE) edit set namesuffix -- "${SUFFIX}"
 	$(KUSTOMIZE) build config/deploy | $(KUBECTL) apply -f -
+	git restore config/manager/pvc.yaml
 
 .PHONY: generate-deploy-manifests
 generate-deploy-manifests: manifests kustomize
+	PVC_SIZE=${PVC_SIZE} envsubst < config/manager/pvc.yaml | sponge config/manager/pvc.yaml
```
piyush-garg, line 228: we need to setup these tools in CI

PuneetPunamiya (approved):

piyush-garg (approved):

PR #164, opened 2026-04-29
Add health endpoint Service and Route for external monitoring

obrienrobert:
Enables external health monitoring to allow endpoint monitoring via Gatus.

### DIFF
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -1,3 +1,4 @@
 resources:
 - manager.yaml
 - pvc.yaml
+- service.yaml
--- a/config/manager/service.yaml
+++ b/config/manager/service.yaml
@@ -1,0 +1,17 @@
+kind: Service
+apiVersion: v1
+metadata:
+  labels:
+    control-plane: controller-manager
+    app.kubernetes.io/name: unstructured-data-controller
+    app.kubernetes.io/managed-by: kustomize
+  name: controller-manager-service
+spec:
+  ports:
+  - name: https
+    protocol: TCP
+    port: 8081
+    targetPort: 8081
+  selector:
+    control-plane: controller-manager
+    app.kubernetes.io/name: unstructured-data-controller


config/default/route.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: route.openshift.io/v1
+kind: Route
+metadata:
+  labels:
+    component: route
+    control-plane: controller-manager
+    app.kubernetes.io/name: unstructured-data-controller
+    app.kubernetes.io/managed-by: kustomize
+    shard: internal
+  name: udc-health
```
piyush-garg: can we rename this to `controller-manager-route`

config/default/kustomization.yaml
```diff
@@ -18,6 +18,8 @@ resources:
 - ../crd
 - ../rbac
 - ../manager
+- service.yaml
```
piyush-garg: it is better to keep the file in `manager` only along with deployment

config/default/kustomization.yaml
piyush-garg, line 1: will it not cause `make deploy` to fail in CI, as CI is on kind and route is openshift thing

config/default/route.yaml
```diff
@@ -0,0 +1,18 @@
...
+spec:
+  port:
+    targetPort: http
```
piyush-garg: can we change it to `https`

config/default/service.yaml
```diff
@@ -0,0 +1,17 @@
...
+apiVersion: v1
+metadata:
+  labels:
+    control-plane: controller-manager
+    app.kubernetes.io/name: unstructured-data-controller
+    app.kubernetes.io/managed-by: kustomize
+  name: controller-manager-service
+spec:
+  ports:
+  - name: http
```
piyush-garg: can we change it to `https`

config/default/route.yaml
obrienrobert: Route name is short to stay under the 63-char DNS label limit. With the unstructured- prefix and long namespace, controller-manager-route produces a 79-char label.

config/default/kustomization.yaml
obrienrobert: route.yaml isn't in any kustomization.yaml so it won't be picked up by make deploy. Kind CI is fine.

piyush-garg (approved):

PR #168, opened 2026-04-30
Refactor HTTP clients to use shared utilities and proper error handling

gshikhar2021:
- The embedding and docling clients had a lot of duplicated code for making HTTP requests - same logic for creating requests, setting headers, handling responses. This was getting messy, especially when we needed to check for specific error codes like rate limits (429) or validation errors (422)

### DIFF
--- a/internal/controller/vectorembeddingsgenerator_controller.go
+++ b/internal/controller/vectorembeddingsgenerator_controller.go
@@ -21,7 +21,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"strings"
 	"time"
 
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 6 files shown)


internal/controller/vectorembeddingsgenerator_controller.go
```diff
@@ -224,12 +224,13 @@ func (r *VectorEmbeddingsGeneratorReconciler) processChunkedFile(ctx context.Con
...
-				logger.Error(err, "embedding API rate limited (429), will retry on next reconciliation", "file", chunksFilePath, "batchStart", batchStart)
-			} else {
-				logger.Error(err, "failed to generate embeddings for batch", "file", chunksFilePath, "batchStart", batchStart, "batchEnd", batchEnd)
-			}
+		if err != nil && commonhttp.IsStatusTooManyRequests(err) {
+			logger.Info("rate limit exceeded (429), will retry after 5 seconds", "batchStart", batchStart, "batchEnd", batchEnd)
+			time.Sleep(5 * time.Second)
+			batchStart -= batchSize
+			continue
+		} else if err != nil {
```
PuneetPunamiya, line 232: I think just returning the error and let the controller's runtime work queue requeue it

```golang
 if err != nil && commonhttp.IsStatusTooManyRequests(err) {
      logger.Info("rate limited (429), will retry on next reconciliation",
          "batchStart", batchStart, "batchEnd", batchEnd)
      return false, err
  } else if err != nil {
      logger.Error(err, "failed to generate embeddings for batch",
          "batchStart", batchStart, "batchEnd", batchEnd)
      return false, err
  }

```

pkg/commonhttp/client.go
concaf, line 1: rename to httputils

PR #202, opened 2026-05-07
Add e2e test for S3 destination processing pipeline

PuneetPunamiya:
* Tests the complete flow from source S3 bucket through document processing, chunking, and vector embeddings generation to destination S3 bucket


test/e2e/unstructured_test.go
```diff
@@ -758,3 +758,420 @@ func TestUnstructuredDataLoad(t *testing.T) {
...
+				Region:          "us-east-1",
+				AccessKeyID:     "test",
+				SecretAccessKey: "test",
+				Endpoint:        localstackURL,
+			})
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			s3Client, err := awsclienthandler.GetSourceS3Client()
```
gshikhar2021: Here we are getting sourceS3client and using it for creating destination bucket at line 844 , it should be destinationS3client ?

test/e2e/unstructured_test.go
```diff
@@ -758,3 +758,420 @@ func TestUnstructuredDataLoad(t *testing.T) {
...
+					DestinationConfig: v1alpha1.DestinationConfig{
+						Type: v1alpha1.TypeS3,
+						S3DestinationConfig: v1alpha1.S3Config{
+							Bucket: destinationBucketName,
+							Prefix: destinationPrefix,
+						},
+					},
+				},
+			}
+
```
gshikhar2021: Unstructured Data Product CR creation is already in utils/utils_function.go , can we pass values as function argument and reuse that instead of creating new one again

test/e2e/unstructured_test.go
```diff
@@ -758,3 +758,420 @@ func TestUnstructuredDataLoad(t *testing.T) {
...
+					key := *obj.Key
+					// embeddings files typically end with .json
+					if strings.HasSuffix(key, ".json") {
+						t.Logf("Found embeddings file: %s", key)
+						foundCount++
+					}
+				}
+
+				// we expect at least as many embeddings files as input files
+				if foundCount >= len(files) {
```
gshikhar2021: Why atleast , I think it should be exact match == , wdyt?

test/e2e/unstructured_test.go
```diff
@@ -758,3 +758,420 @@ func TestUnstructuredDataLoad(t *testing.T) {
...
+		// verify embeddings files exist in destination bucket
+		t.Log("Verifying embeddings files in destination S3 bucket...")
+
+		destS3Client, err := awsclienthandler.GetDestinationS3Client()
+		if err != nil {
+			t.Error(err)
+		}
+
+		// poll until embeddings files are found in destination bucket
+		embeddingsFound := false
```
gshikhar2021: do we really need extra variable as embedding found, what I think if 
- it finds all the files (==) condition satisfied then automatically it will return true 
- if it gets timeout then it is for sure it won't be able to find files so directly log error , instead of adding extra check at line 1096-1098

test/e2e/unstructured_test.go
```diff
@@ -758,3 +758,420 @@ func TestUnstructuredDataLoad(t *testing.T) {
...
+				Bucket: aws.String(destinationBucketName),
+			})
+			if err == nil {
+				for _, obj := range destOutput.Contents {
+					_, _ = s3Client.DeleteObject(ctx, &s3.DeleteObjectInput{
+						Bucket: aws.String(destinationBucketName),
+						Key:    obj.Key,
+					})
+				}
+			}
```
gshikhar2021: Here 2 things, 
- either we remove all these because when we are deleting the buckets then why do we care about clearing all the objects , it will be automatically cleared
- even if we are sticking with this approach then we are using sourceS3client to list objects from destination bucket [line 1150-1152]

test/e2e/unstructured_test.go
PuneetPunamiya: updated!

test/e2e/unstructured_test.go
PuneetPunamiya: updated!

test/e2e/unstructured_test.go
```diff
@@ -759,3 +759,420 @@ func TestUnstructuredDataLoad(t *testing.T) {
...
+				})
+				if err == nil {
+					for _, obj := range destOutput.Contents {
+						_, _ = destS3Client.DeleteObject(ctx, &s3.DeleteObjectInput{
+							Bucket: aws.String(destinationBucketName),
+							Key:    obj.Key,
+						})
+					}
+				}
+			}
```
gshikhar2021: can we remove these lines because at the end of the day we are deleting buckets ?

test/e2e/unstructured_test.go
```diff
@@ -767,3 +767,420 @@ func TestUnstructuredDataLoad(t *testing.T) {
...
+	dataProductCRName := "testunstructureddataproduct"
+	sourcePrefix := "testunstructureddataproduct/source-data"
+	destinationPrefix := "testunstructureddataproduct/processed-data"
+
+	var queueURL string
+	testFilesDirectory := "test/resources/unstructured/unstructured-files"
+
+	var kubeClient klient.Client
+
+	feature.Setup(
```
piyush-garg: why we are not reusing the feature.setup

test/e2e/unstructured_test.go
```diff
@@ -738,6 +747,428 @@ func TestUnstructuredDataLoad(t *testing.T) {
...
+			SecretAccessKey: "test",
+			Endpoint:        localstackURL,
+		}
+		if err := awsclienthandler.NewSourceS3ClientFromConfig(ctx, e2eAWS); err != nil {
+			t.Fatal(err)
+		}
+		if err := awsclienthandler.NewDestinationS3ClientFromConfig(ctx, e2eAWS); err != nil {
+			t.Fatal(err)
+		}
+		sourceS3Client, err := awsclienthandler.GetSourceS3Client()
```
piyush-garg: why we need to create these client in every test ?

test/e2e/unstructured_test.go
```diff
@@ -738,6 +747,428 @@ func TestUnstructuredDataLoad(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		destS3Client, err := awsclienthandler.GetDestinationS3Client()
```
piyush-garg: why these clients are getting created again and again ?

test/e2e/unstructured_test.go
```diff
@@ -738,6 +747,428 @@ func TestUnstructuredDataLoad(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		t.Log("upload file1 to source bucket")
```
piyush-garg: there should be file already because of first test

test/e2e/unstructured_test.go
```diff
@@ -66,6 +66,15 @@ func TestUnstructuredDataLoad(t *testing.T) {
 
 	queueURL := "http://sqs.us-east-1.localhost.localstack.cloud:4566/000000000000/" + unstructuredQueueName
 	unstructuredFilesDirectory := "test/resources/unstructured/unstructured-files"
+	destinationBucketName := "result-bucket"
+	destinationPrefix := schemaName + "/processed-data"
+
+	hashTestFile1Key := fmt.Sprintf("%s/e2e-hash/file1.pdf", schemaName)
+	hashTestFile2Key := fmt.Sprintf("%s/e2e-hash/file2.pdf", schemaName)
+
+	var file1DestKey string
```
piyush-garg: why global variables

test/e2e/unstructured_test.go
```diff
@@ -66,6 +66,15 @@ func TestUnstructuredDataLoad(t *testing.T) {
 
 	queueURL := "http://sqs.us-east-1.localhost.localstack.cloud:4566/000000000000/" + unstructuredQueueName
 	unstructuredFilesDirectory := "test/resources/unstructured/unstructured-files"
+	destinationBucketName := "result-bucket"
+	destinationPrefix := schemaName + "/processed-data"
+
+	hashTestFile1Key := fmt.Sprintf("%s/e2e-hash/file1.pdf", schemaName)
```
piyush-garg: why new files ?

test/e2e/unstructured_test.go
```diff
@@ -738,6 +747,428 @@ func TestUnstructuredDataLoad(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		destS3Client, err := awsclienthandler.GetDestinationS3Client()
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		_, err = destS3Client.CreateBucket(ctx, &s3.CreateBucketInput{
```
piyush-garg: why new bucket ?

test/e2e/unstructured_test.go
```diff
@@ -738,6 +747,428 @@ func TestUnstructuredDataLoad(t *testing.T) {
...
+			v1alpha1.UnstructuredDataPipelineCondition,
+			"unstructureddatapipelines.operator.dataverse.redhat.com",
+			dataProductCRName,
+			testNamespace,
+		); err != nil {
+			t.Fatal(err)
+		}
+		t.Log("UnstructuredDataPipeline is ready after S3 destination patch")
+
+		if err := apimachinerywait.PollUntilContextTimeout(
```
piyush-garg: why this wait is required ?

test/e2e/unstructured_test.go
```diff
@@ -738,6 +747,428 @@ func TestUnstructuredDataLoad(t *testing.T) {
...
+				}
+				t.Log("file1 embedding not in destination yet, retrying...")
+				return false, nil
+			},
+		)
+		if err != nil {
+			t.Fatal("timeout waiting for file1 embedding in destination bucket")
+		}
+
+		head1, err := destS3Client.HeadObject(ctx, &s3.HeadObjectInput{
```
piyush-garg: this should be a utility function

test/e2e/unstructured_test.go
PuneetPunamiya: updated!

test/e2e/unstructured_test.go
PuneetPunamiya: updated!

test/e2e/unstructured_test.go
PuneetPunamiya: updated!

PR #209, opened 2026-05-13
Fix: Remove logging of S3 presigned URL

gshikhar2021:
- Removing the log statement printing s3 presigned URL in plain text for security enhancement

### DIFF
--- a/pkg/docling/client.go
+++ b/pkg/docling/client.go
@@ -240,7 +240,13 @@ func (c *Client) ConvertFile(
 		return nil, fmt.Errorf("failed to marshal config to docling payload: %w", err)
 	}
 
-	logger.Info("sending request to convert file", "url", convertSourceAsyncEndpoint, "http source", fileURL)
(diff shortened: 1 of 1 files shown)


pkg/docling/client.go
```diff
@@ -240,7 +240,7 @@ func (c *Client) ConvertFile(
 		return nil, fmt.Errorf("failed to marshal config to docling payload: %w", err)
 	}
 
-	logger.Info("sending request to convert file", "url", convertSourceAsyncEndpoint, "http source", fileURL)
+	logger.Info("sending request to convert file", "url", convertSourceAsyncEndpoint)
```
piyush-garg: should we add some logs to determine which file it is trying, we dont need to log presigned url, but some identifier

gshikhar2021: added rawFile Path

piyush-garg (approved):

piyush-garg (approved):

PR #232, opened 2026-05-28
Support fetching of files/permissions from gdrive

vinamra28:
Given a list of folder IDs from google drive, we should be able to do a recursive crawl and fetch the files along with their permissions.

For each file, following metadata is fetched:

- file_id
- file_name
- mime_type
- created_at
- updated_at
- owner
- file_size
- root_folder_id
- parent_folder_id
- parent_folder_name
- status (whether fetching of metadata for the file was a failure or success)

This package outputs the following:

- crawl_metadata.json which contains the list of metadata for each file
- inaccessible.json which contains inaccessible_folders, inaccessible_files, inaccessible_shortcut_target_folders, inaccessible_shortcut_target_files
- permissions for each file ID

Fetching of permissions is done as following:

- fetch the file permissions
- if the fetched entity is a group, it should be resolved using ldap or google groups
- Once the group members are fetched, the same is cached for future reference

The service account used here should have write/edit privileges on the folder it's crawling so that it can also lookup for the permissions.


pkg/ldap/client.go
```diff
@@ -0,0 +1,161 @@
...
+	// UIDAttribute is the LDAP attribute used as the unique user identifier
+	// (e.g., "uid"). Used by ResolveUser to extract the UID from search results.
+	UIDAttribute string `yaml:"uidAttribute"`
+
+	// Attributes is the list of LDAP attributes to retrieve in queries.
+	Attributes []string `yaml:"attributes"`
+
+	// DialTimeout is the connection timeout. Defaults to 5 seconds if zero.
+	DialTimeout time.Duration `yaml:"dialTimeout"`
+}
```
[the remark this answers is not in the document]
vinamra28: Intentionally keeping anonymous bind only for this PR — the pathfinder reference code (which is battle-tested in production) only uses `UnauthenticatedBind`. Adding `BindDN`/`BindPassword` support is a valid enhancement but out of scope for the initial gdrive package. We can add it in a follow-up PR when there's a concrete need.

pkg/ldap/client.go
```diff
@@ -0,0 +1,161 @@
...
+	server           string
+	groupDN          string
+	userDN           string
+	baseUserDN       string
+	userSearchFilter string
+	emailAttribute   string
+	uidAttribute     string
+	attributes       []string
+	dialTimeout      time.Duration
+}
```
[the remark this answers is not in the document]
vinamra28: Same as above — `bindDN`/`bindPassword` fields are out of scope for this PR. The pathfinder reference only uses anonymous bind. Will address in a follow-up if authenticated bind is needed.

pkg/ldap/client.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+		err = newConn.UnauthenticatedBind("")
+		if err != nil {
+			_ = newConn.Close()
+			return nil
+		}
+		l.conn = newConn
+	}
+	return l.conn
+}
```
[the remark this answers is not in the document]
vinamra28: The nil `conn` case cannot occur in normal usage — `InitLdap` always initializes the connection, and `getConn` only replaces it on `IsClosing()`. Adding lazy initialization changes the semantics and isn't needed. The pathfinder reference code has the same behavior. The authenticated bind suggestion is addressed in the other comment (out of scope for this PR).

pkg/ldap/client.go
```diff
@@ -0,0 +1,161 @@
...
+		groupDN:          config.GroupDN,
+		userDN:           config.UserDN,
+		baseUserDN:       config.BaseUserDN,
+		userSearchFilter: config.UserSearchFilter,
+		emailAttribute:   config.EmailAttribute,
+		uidAttribute:     config.UIDAttribute,
+		attributes:       config.Attributes,
+		dialTimeout:      config.DialTimeout,
+	}, nil
+}
```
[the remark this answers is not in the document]
vinamra28: Partially addressed — added `ensureAttribute` helper that auto-appends `EmailAttribute` and `UIDAttribute` to the `Attributes` list if missing. The `BindDN`/`BindPassword` support is intentionally deferred (see other replies).

pkg/gdrive/crawler.go
```diff
@@ -0,0 +1,484 @@
...
+
+	logger.Info("starting folder crawl", "rootFolderName", rootFolder.Name)
+
+	skipSet := make(map[string]bool, len(c.config.SkipFolderNames))
+	for _, name := range c.config.SkipFolderNames {
+		skipSet[name] = true
+	}
+
+	result := &CrawlResult{}
+	visitedFolders := &sync.Map{}
```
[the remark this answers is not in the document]
vinamra28: Intentionally keeping single-threaded traversal within a root folder — the pathfinder reference code (production-proven) uses the same synchronous recursive approach. Concurrent subfolder crawling would be a significant architectural change requiring careful handling of the visited-folders map, result ordering, and API rate limiting. The existing `CrawlFolders` already provides concurrency across root folders. Concurrent subfolder crawling can be considered as a performance optimization in a follow-up PR if needed.

api/v1alpha1/controllerconfig_types.go
```diff
@@ -50,6 +50,24 @@ type ControllerConfigSpec struct {
 	DataStorageBucket           string `json:"dataStorageBucket,omitempty"`
 	// +kubebuilder:validation:Minimum=1
 	UnstructuredDataPipelineResyncInterval *int `json:"unstructuredDataPipelineResyncInterval,omitempty"`
+	// GDriveConfig holds operator-level Google Drive crawling settings.
+	// +optional
+	GDriveConfig *GDriveControllerConfig `json:"gdriveConfig,omitempty"`
```
concaf:
```suggestion
	GoogleDriveConfig *GoogleDriveControllerConfig `json:"googleDriveConfig,omitempty"`
```

api/v1alpha1/unstructureddatapipeline_types.go
```diff
@@ -140,8 +141,44 @@ type PipelineStage struct {
...
+	// FolderIDs is the list of Google Drive folder IDs to crawl recursively.
+	// +kubebuilder:validation:MinItems=1
+	FolderIDs []string `json:"folderIDs"`
+	// SkipFolderNames is an optional list of folder names to skip during crawling.
+	// +optional
+	SkipFolderNames []string `json:"skipFolderNames,omitempty"`
+}
+
+// LDAPConfig holds configuration for connecting to an LDAP server for identity resolution.
+type LDAPConfig struct {
```
concaf: the struct exists here but is not used here

config/samples/unstructured-secret.yaml
```diff
@@ -24,3 +24,22 @@ stringData:
   AWS_ACCESS_KEY_ID: test
   AWS_SECRET_ACCESS_KEY: test
   AWS_ENDPOINT: http://localhost:4566
+---
+# Pipeline-level secret for Google Drive credentials
+apiVersion: v1
+kind: Secret
+metadata:
+  name: gdrive-pipeline-secret
```
concaf: we should have one secret per pipeline

concaf (changes requested):

internal/controller/controllerutils/secret_helpers.go
```diff
@@ -54,3 +55,21 @@ func AWSConfigFromSecret(ctx context.Context, c client.Client, secretName, names
 		Endpoint:        string(secret.Data["AWS_ENDPOINT"]),
 	}, nil
 }
+
+// GDriveCredentialsFromSecret reads the Google service account JSON
+// from a K8s Secret. The secret must contain a key named
+// "GOOGLE_SERVICE_ACCOUNT_JSON".
+func GDriveCredentialsFromSecret(ctx context.Context, c client.Client, secretName, namespace string) ([]byte, error) {
```
concaf, line 62: we should use one secret

internal/controller/config_utils.go
```diff
@@ -25,6 +25,12 @@ func IsAWSClientNotInitializedError(err error) bool {
 		strings.Contains(msg, errMsgPresignNotInitialized)
 }
 
+func GetGDriveControllerConfig() *operatorv1alpha1.GDriveControllerConfig {
```
concaf: why is this needed ?

.github/workflows/test-e2e.yml
```diff
@@ -150,6 +150,13 @@ jobs:
       - name: Install sponge
         run: sudo apt-get update && sudo apt-get install -y moreutils
 
+      - name: E2E tests without Kind
+        run: |
+          service_account=$(echo ${{ secrets.GDRIVE_SERVICE_ACCOUNT_ENCODED }} | base64 -d)
+          echo "$service_account" > unstructured-data-controller-24ad7a29c246.json
+          export GDRIVE_SERVICE_ACCOUNT_FILE=$(pwd)/unstructured-data-controller-24ad7a29c246.json
+          go test -tags=e2e -v -count=1 ./test/pkg/...
```
concaf: Why is this needed? Why do we need to run end to end tests without kind?

api/v1alpha1/controllerconfig_types.go
```diff
@@ -50,6 +50,24 @@ type ControllerConfigSpec struct {
...
+	// +optional
+	MaxRetries int `json:"maxRetries,omitempty"`
+	// ConcurrentFolders is the maximum number of folders to crawl concurrently. Defaults to 5.
+	// +optional
+	ConcurrentFolders int `json:"concurrentFolders,omitempty"`
+	// ConcurrentDownloads is the maximum number of files to download concurrently. Defaults to 10.
+	// +optional
+	ConcurrentDownloads int `json:"concurrentDownloads,omitempty"`
+	// LDAPConfig holds LDAP configuration for user/group identity resolution.
+	LDAPConfig LDAPConfig `json:"ldapConfig"`
```
concaf: Thinking out loud, should LDAP config be at the root level or should it be nested inside Google Drive controller config?

api/v1alpha1/unstructureddatapipeline_types.go
```diff
@@ -73,6 +73,7 @@ type (
 
 const (
 	TypeS3                             UnstructuredDataType = "s3"
+	TypeGDrive                         UnstructuredDataType = "gdrive"
```
concaf:
```suggestion
	TypeGoogleDrive                         UnstructuredDataType = "googleDrive"
```

api/v1alpha1/unstructureddatapipeline_types.go
```diff
@@ -140,8 +141,44 @@ type PipelineStage struct {
...
-	Type     UnstructuredDataType `json:"type,omitempty"`
-	S3Config S3Config             `json:"s3Config,omitempty"`
+	Type         UnstructuredDataType `json:"type,omitempty"`
+	S3Config     S3Config             `json:"s3Config,omitempty"`
+	GDriveConfig *GDriveConfig        `json:"gdriveConfig,omitempty"`
+}
+
+// GDriveConfig configures Google Drive folder crawling at the pipeline level.
+// Controller-level settings (maxRetries, concurrency, LDAP) are in ControllerConfig.
+type GDriveConfig struct {
```
concaf:
```suggestion
type GoogleDriveConfig struct {
```

api/v1alpha1/unstructureddatapipeline_types.go
```diff
@@ -140,8 +141,44 @@ type PipelineStage struct {
...
+	// FolderIDs is the list of Google Drive folder IDs to crawl recursively.
+	// +kubebuilder:validation:MinItems=1
+	FolderIDs []string `json:"folderIDs"`
```
concaf:
```suggestion
type GoogleDriveFolder struct {
  ID string
}

	Folders []GoogleDriveFolder `json:"folders"`
```

config/samples/operator_v1alpha1_unstructureddatapipeline_gdrive.yaml
```diff
@@ -0,0 +1,55 @@
...
+    - name: crawl
+      type: SourceCrawler
+      sourceCrawlerConfig:
+        type: gdrive
+        gdriveConfig:
+          folderIDs:
+            - "1ua4tWAwVVZV_0p93asX5KJt7ycKH03ta"
+          skipFolderNames:
+            - "_indexes"
+            - "_synthesized"
```
concaf:
```suggestion
          folders:
            - id: "1ua4tWAwVVZV_0p93asX5KJt7ycKH03ta"
          skipFolders:
            - pattern: "_indexes"
            - pattern: "_synthesized"
```

internal/controller/controllerconfig_controller.go
```diff
@@ -115,6 +118,15 @@ func (r *ControllerConfigReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
 	}
 
+	gdriveConfigMu.Lock()
+	if config.Spec.GDriveConfig != nil {
+		gdriveControllerConfig = config.Spec.GDriveConfig.DeepCopy()
+		logger.Info("gdrive controller config loaded")
+	} else {
+		gdriveControllerConfig = nil
+	}
+	gdriveConfigMu.Unlock()
```
concaf: why is this needed ?

internal/controller/sourcecrawler_controller.go
```diff
@@ -176,6 +185,85 @@ func handleSQSWakeUp(ctx context.Context, queueURL, bucket, prefix string) (ctrl
...
+		GroupDN:          ctrlConfig.LDAPConfig.GroupDN,
+		UserDN:           ctrlConfig.LDAPConfig.UserDN,
+		BaseUserDN:       ctrlConfig.LDAPConfig.BaseUserDN,
+		UserSearchFilter: ctrlConfig.LDAPConfig.UserSearchFilter,
+		EmailAttribute:   ctrlConfig.LDAPConfig.EmailAttribute,
+		Attributes:       ctrlConfig.LDAPConfig.Attributes,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize LDAP client: %w", err)
+	}
```
concaf: should be initialized in controller config ?

pkg/unstructured/source.go
```diff
@@ -224,3 +228,267 @@ func (s *S3BucketSource) s3Key(filestorePath string) string {
...
+				mu.Lock()
+				storedFiles = append(storedFiles, file)
+				mu.Unlock()
+			}
+			return nil
+		})
+	}
+	_ = dlGroup.Wait()
+
+	// Phase 3: Garbage collection — delete files and permissions no longer in source
```
concaf, line 347: let's not delete files for now ?

concaf (changes requested):

pkg/unstructured/file.go
```diff
@@ -45,6 +45,10 @@ func MetadataPath(rawFilePath string) string {
 	return rawFilePath + ".json"
 }
 
+func PermissionsPath(rawFilePath string) string {
+	return rawFilePath + ".permissions.json"
+}
```
concaf, line 50: We should store it as part of regular metadata.json, right?

internal/controller/controllerutils/secret_helpers.go
vinamra28: it's a helper function which extracts the google service account from the secret, not able to understand the concern here 😅

pkg/unstructured/file.go
vinamra28: storing it separately becomes easy while parsing in snowflake external tables/materialized views.

api/v1alpha1/controllerconfig_types.go
```diff
@@ -30,6 +30,16 @@ import (
 //	  maxConcurrentDoclingTasks: 5
```
concaf, line 30: `stages/crawl/permissions/fileid.json` <-- let's use this name

api/v1alpha1/sourcecrawler_types.go
```diff
@@ -20,16 +20,28 @@ import (
...
+//
+// sample spec (Google Drive):
+//
+//	spec:
+//	  stageName: crawl
+//	  sourceCrawlerConfig:
+//	    type: googleDrive
+//	    googleDriveConfig:
+//	      folders:
+//	        - id: "1ABCdef_example_folder_id"
```
concaf: we should take `url` here

internal/controller/sourcecrawler_controller.go
```diff
@@ -135,58 +148,132 @@ func (r *SourceCrawlerReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+) (*unstructured.GDriveSource, error) {
+	if gdriveConfig == nil {
+		return nil, errors.New("gdriveConfig is required when source type is gdrive")
+	}
+
+	controllerConfigCR, err := GetControllerConfig(ctx, r.Client, sourceCrawlerCR.Namespace)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get ControllerConfig: %w", err)
+	}
+	ctrlConfig := controllerConfigCR.Spec.GoogleDriveConfig
```
concaf: this should be a global variable and not be extracted from controller config in run time

concaf (changes requested):

internal/controller/config_utils.go
```diff
@@ -25,6 +28,25 @@ func IsAWSClientNotInitializedError(err error) bool {
...
 
+func GetGDriveLDAPClient() ldap.Client {
+	return LDAPClient
+}
+
+func GetGDriveCacheClient() pkgcache.Cache {
+	return CacheClient
+}
+
+func GetControllerConfig(ctx context.Context, kubeClient client.Client, namespace string) (*operatorv1alpha1.ControllerConfig, error) {
```
concaf: not needed imo

pkg/unstructured/source.go
```diff
@@ -224,3 +228,270 @@ func (s *S3BucketSource) s3Key(filestorePath string) string {
...
+	currentFiles := make(map[string]string, len(fileRecords))
+
+	dlGroup, _ := errgroup.WithContext(ctx)
+	dlGroup.SetLimit(g.ConcurrentDownloads)
+
+	for _, record := range fileRecords {
+		ext := path.Ext(record.FileName)
+		if strings.HasPrefix(record.MimeType, "application/vnd.google-apps.") {
+			ext = ".pdf"
+		}
```
concaf, line 316: Why is everything downloaded as a PDF? Why can we not download in the native format so that Dockling and other document conversion can handle based on the type of file that would be ideal?

pkg/unstructured/source.go
```diff
@@ -224,3 +228,270 @@ func (s *S3BucketSource) s3Key(filestorePath string) string {
...
+				mu.Unlock()
+				return nil
+			}
+			if stored {
+				logger.Info("stored gdrive file",
+					"fileID", record.FileID, "fileName", record.FileName)
+				mu.Lock()
+				storedFiles = append(storedFiles, file)
+				mu.Unlock()
+			}
```
concaf, line 341: do we need this ? as filestore already provides thread safe storage

vinamra28: The mu.Lock() is needed because storedFiles is appended to from multiple goroutines concurrently - the dlGroup.Go calls run up to ConcurrentDownloads (default 10) goroutines in parallel. Without the mutex, concurrent appends to the same slice would be a data race.

api/v1alpha1/controllerconfig_types.go
vinamra28: addressed

api/v1alpha1/sourcecrawler_types.go
vinamra28: addressed

internal/controller/sourcecrawler_controller.go
vinamra28: addressed

internal/controller/config_utils.go
vinamra28: addressed

concaf (approved):

PR #241, opened 2026-05-29
Add stages-based pipeline DAG orchestration

concaf:
## Summary

- Refactors `UnstructuredDataProduct` from a hardcoded 3-stage pipeline to a DAG-based `stages` model where each stage declares its type, config, and dependencies
- Adds 2 new CRDs (`Crawler`, `DestinationSync`) with controllers, extracted from the pipeline controller
- Pipeline controller becomes a pure DAG orchestrator — creates child CRs, monitors completion, triggers next stages
- All child controllers now trigger the pipeline back via force-reconcile instead of hardcoding their successor
- Adds DAG validation utilities (cycle detection, dependency resolution) with 17 unit tests
- Lays groundwork for Phase 2: `type: Custom` with `customConfig.podSpec` for bring-your-own-container stages

### Target spec shape

```yaml
spec:
  stages:
    - name: crawl
      type: Crawler
      crawlerConfig:
        sourceConfig: { type: s3, s3Config: { bucket: my-bucket } }
    - name: convert
      depends: [crawl]
      type: DocumentProcessor
      documentProcessorConfig: { ... }
    - name: chunk
      depends: [convert]
      type: ChunksGenerator
      chunksGeneratorConfig: { ... }
    - name: embed
      depends: [chunk]
      type: VectorEmbeddingsGenerator
      vectorEmbeddingsGeneratorConfig: { ... }
    - name: sync
      depends: [embed]
      type: DestinationSync
      destinationSyncConfig:
        destinationConfig: { ... }
```

## Breaking Changes

- `UnstructuredDataProductSpec` no longer has top-level `sourceConfig`, `destinationConfig`, `documentProcessorConfig`, `chunksGeneratorConfig`, `vectorEmbeddingsGeneratorConfig` fields — replaced by `stages` array
- Child CRs are now named `<pipelineName>-<stageName>` instead of `<pipelineName>` — existing child CRs will be orphaned and need manual cleanup

## Test plan

- [x] `make build` compiles
- [x] `make lint` passes (0 issues)
- [x] `make test` passes (17 DAG unit tests + existing tests)
- [ ] E2E: deploy with stages-format pipeline CR, verify full pipeline flow
- [ ] Migration: update existing pipeline CRs to stages format before operator upgrade


api/v1alpha1/unstructureddatapipeline_types.go
```diff
@@ -20,6 +20,54 @@ import (
...
+//	      sourceCrawlerConfig:
+//	        sourceConfig:
+//	          type: s3
+//	          s3Config:
+//	            bucket: data-ingestion-bucket
+//	            prefix: documents/
+//	            secretRef: source-aws-creds   # k8s secret with AWS credentials
+//	    - name: convert
+//	      type: DocumentProcessor
+//	      depends: [crawl]
```
concaf: change this to 

```
depends:
- name: crawl
```

api/v1alpha1/unstructureddatapipeline_types.go
```diff
@@ -20,6 +20,54 @@ import (
...
 )
 
+// Sample spec:
+//
+//	spec:
+//	  stages:
+//	    - name: crawl
+//	      type: SourceCrawler
+//	      sourceCrawlerConfig:
+//	        sourceConfig:
```
PuneetPunamiya: Can we remove this ?

api/v1alpha1/unstructureddatapipeline_types.go
```diff
@@ -35,6 +83,63 @@ const (
...
+	// +optional
+	ChunksGeneratorConfig *ChunksGeneratorConfig `json:"chunksGeneratorConfig,omitempty"`
+	// +optional
+	VectorEmbeddingsGeneratorConfig *VectorEmbeddingsGeneratorConfig `json:"vectorEmbeddingsGeneratorConfig,omitempty"`
+	// +optional
+	DestinationSyncerConfig *DestinationSyncerStageConfig `json:"destinationSyncConfig,omitempty"`
+}
+
+// SourceCrawlerStageConfig wraps SourceConfig for the Crawler stage type.
+type SourceCrawlerStageConfig struct {
```
PuneetPunamiya: Can we remove this struct

api/v1alpha1/unstructureddatapipeline_types.go
PuneetPunamiya:
```suggestion
//	        type:
```

api/v1alpha1/unstructureddatapipeline_types.go
```diff
@@ -88,6 +194,7 @@ type TokenSplitterConfig struct {
 type VectorEmbeddingsGeneratorConfig struct {
 	ModelName               string                  `json:"modelName,omitempty"`
 	NomicEmbedTextV15Config NomicEmbedTextV15Config `json:"nomicEmbedTextV15Config,omitempty"`
+	SecretRef               string                  `json:"secretRef,omitempty"`
```
PuneetPunamiya: I would recommend to have all secrets in one unified secret

api/v1alpha1/unstructureddatapipeline_types.go
```diff
@@ -96,43 +203,44 @@ type NomicEmbedTextV15Config struct {
...
 }
 
 // S3Config configures an S3 bucket and optional prefix.
 type S3Config struct {
-	Bucket string `json:"bucket"`
-	Prefix string `json:"prefix,omitempty"`
+	Bucket      string `json:"bucket"`
+	Prefix      string `json:"prefix,omitempty"`
+	SQSQueueURL string `json:"sqsQueueURL,omitempty"`
+	SecretRef   string `json:"secretRef,omitempty"`
```
PuneetPunamiya: Move this to unified secret

api/v1alpha1/chunksgenerator_types.go
```diff
@@ -20,13 +20,34 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// Sample spec:
+//
+//	spec:
+//	  pipelineName: my-pipeline
```
PuneetPunamiya: Why do we need this ?

api/v1alpha1/documentprocessor_types.go
```diff
@@ -22,13 +22,37 @@ import (
...
+//	  jobs: []                               # tracks in-flight docling conversion jobs
+
 const (
 	DocumentProcessorCondition = "DocumentProcessorReady"
 )
 
 // DocumentProcessorSpec defines the desired state of DocumentProcessor
 type DocumentProcessorSpec struct {
-	DataProduct             string                  `json:"dataProduct,omitempty"`
+	DataPipeline            string                  `json:"pipelineName,omitempty"`
```
PuneetPunamiya: Can we remove this ?

api/v1alpha1/vectorembeddingsgenerator_types.go
```diff
@@ -22,13 +22,31 @@ import (
...
+//	      message: successfully reconciled
+
 const (
 	VectorEmbeddingGenerationConditionType = "VectorEmbeddingGenerationReady"
 )
 
 // VectorEmbeddingsGeneratorSpec defines the desired state of VectorEmbeddingsGenerator.
 type VectorEmbeddingsGeneratorSpec struct {
-	DataProduct                     string                          `json:"dataProduct,omitempty"`
+	DataPipeline                    string                          `json:"pipelineName,omitempty"`
```
PuneetPunamiya: Can we remove this ?

api/v1alpha1/sourcecrawler_types.go
```diff
@@ -0,0 +1,126 @@
...
+//	      status: "True"                     # True | False | Unknown
+//	      message: successfully reconciled
+
+const (
+	SourceCrawlerCondition = "SourceCrawlerReady"
+)
+
+// SourceCrawlerSpec defines the desired state of SourceCrawler.
+type SourceCrawlerSpec struct {
+	DataPipeline             string                   `json:"pipelineName,omitempty"`
```
PuneetPunamiya: Can we remove this ?

api/v1alpha1/unstructureddatapipeline_types.go
```diff
@@ -20,21 +20,123 @@ import (
...
+//	          bucket: data-ingestion-bucket
+//	          prefix: documents/
+//	    - name: convert
+//	      type: DocumentProcessor
+//	      depends: [crawl]
+//	      documentProcessorConfig:
+//	        type: docling
+//	    - name: chunk
+//	      type: ChunksGenerator
+//	      depends: [convert]
```
PuneetPunamiya: We should add the validations like 
Convert should always have dependency as crawler 
Chunks should always have dependecny as documentProcessor
Embeddings should always have dependency on chunks

internal/controller/unstructureddatapipeline_controller.go
```diff
@@ -50,310 +46,348 @@ var (
...
+			})
+		}
 		return r.Status().Update(ctx, latest)
-	}); err != nil {
-		logger.Error(err, "failed to update UnstructuredDataPipeline CR status")
-		return ctrl.Result{}, err
+	})
+}
+
+func (r *UnstructuredDataPipelineReconciler) deleteOrphanedStageCRs(ctx context.Context, stages []operatorv1alpha1.PipelineStage, pipeline *operatorv1alpha1.UnstructuredDataPipeline) error {
```
PuneetPunamiya, line 148: This functions seems to be very complex, can you simplify this function
Basically like get the current CR and the expected new CR diff and update the CR

internal/controller/documentprocessor_controller.go
```diff
@@ -532,12 +530,34 @@ func (r *DocumentProcessorReconciler) getFileUID(ctx context.Context, rawFilePat
...
+	return requests
+}
+
 // SetupWithManager sets up the controller with the Manager.
 func (r *DocumentProcessorReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	labelPredicate := controllerutils.ForceReconcilePredicate()
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&operatorv1alpha1.DocumentProcessor{}).
-		WithEventFilter(predicate.Or(predicate.GenerationChangedPredicate{}, labelPredicate)).
+		For(&operatorv1alpha1.DocumentProcessor{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
```
PuneetPunamiya, line 508: Can you add comments to explain

api/v1alpha1/unstructureddatapipeline_types.go
concaf: Consider future scenarios:
- A DocumentProcessor that reads from another DocumentProcessor's output (multi-pass conversion)
- A ChunksGenerator that reads directly from a SourceCrawler (if the source provides pre-converted text)
- Custom stages (future BYO) that we can't validate against at compile time

pkg/unstructured/file.go
```diff
@@ -17,10 +17,22 @@ limitations under the License.
 package unstructured
 
 import (
+	"path/filepath"
 	"slices"
 	"strings"
 )
 
+func StagePath(pipelineName, stageName string) string {
+	return "pipelines/" + pipelineName + "/stages/" + stageName + "/"
```
PuneetPunamiya: This will change I think, because we are removing `PipelineName` field right ?

internal/controller/sourcecrawler_controller.go
```diff
@@ -0,0 +1,255 @@
...
+// SourceCrawlerReconciler reconciles a SourceCrawler object
+type SourceCrawlerReconciler struct {
+	client.Client
+	Scheme *runtime.Scheme
+}
+
+// +kubebuilder:rbac:groups=operator.dataverse.redhat.com,namespace=unstructured-controller-namespace,resources=sourcecrawlers,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=operator.dataverse.redhat.com,namespace=unstructured-controller-namespace,resources=sourcecrawlers/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=operator.dataverse.redhat.com,namespace=unstructured-controller-namespace,resources=sourcecrawlers/finalizers,verbs=update
+
```
PuneetPunamiya, line 58: We might have to add a marker to watch for secrets also 
// +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch

internal/controller/sourcecrawler_controller.go
```diff
@@ -0,0 +1,255 @@
...
+// SetupWithManager sets up the controller with the Manager.
+func (r *SourceCrawlerReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&operatorv1alpha1.SourceCrawler{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Watches(&operatorv1alpha1.DocumentProcessor{}, handler.EnqueueRequestsFromMapFunc(r.findDependents)).
+		Watches(&operatorv1alpha1.ChunksGenerator{}, handler.EnqueueRequestsFromMapFunc(r.findDependents)).
+		Watches(&operatorv1alpha1.VectorEmbeddingsGenerator{}, handler.EnqueueRequestsFromMapFunc(r.findDependents)).
+		Watches(&operatorv1alpha1.DestinationSyncer{}, handler.EnqueueRequestsFromMapFunc(r.findDependents)).
+		Watches(&corev1.Secret{}, handler.EnqueueRequestsFromMapFunc(r.findSecretDependents)).
+		Named("sourcecrawler").
```
PuneetPunamiya, line 258: SourceCrawler is usually the first stage. Watching DocumentProcessor, ChunksGenerator, VectorEmbeddingsGenerator, and DestinationSyncer implies a crawler can depend on downstream stages, which is uncommon

internal/controller/sourcecrawler_controller.go
```diff
@@ -0,0 +1,255 @@
...
+			Prefix:    sourceCrawlerConfig.S3Config.Prefix,
+			OutputDir: outputDir,
+		}
+	default:
+		return r.handleError(ctx, sourceCrawlerCR, fmt.Errorf("unsupported source type: %s", sourceCrawlerConfig.Type))
+	}
+
+	storedFiles, err := source.SyncFilesToFilestore(ctx, fs)
+	if err != nil {
+		logger.Error(err, "failed to store files to filestore")
```
PuneetPunamiya: The error will logged twice

internal/controller/sourcecrawler_controller.go
```diff
@@ -0,0 +1,255 @@
...
+
+	sourceCrawlerCR := &operatorv1alpha1.SourceCrawler{}
+	if err := r.Get(ctx, req.NamespacedName, sourceCrawlerCR); err != nil {
+		logger.Error(err, "failed to get SourceCrawler CR")
+		return ctrl.Result{}, err
+	}
+
+	if len(sourceCrawlerCR.Spec.Depends) > 0 {
+		allReady, err := controllerutils.CheckDependenciesExist(ctx, r.Client, controllerutils.PipelineNameFromOwner(sourceCrawlerCR), sourceCrawlerCR.Namespace, sourceCrawlerCR.Spec.Depends)
+		if err != nil {
```
PuneetPunamiya: `return ctrl.Result{}, nil` means we never retry unless a watch wakes us up. Please add RequeueAfter: 10 * time.Second here as a fallback so we don't get stuck waiting for upstream CRs

internal/controller/destinationsyncer_controller.go
```diff
@@ -0,0 +1,238 @@
...
+		return r.handleError(ctx, destinationSyncCR, fmt.Errorf("unsupported destination type: %s", destConfig.Type))
+	}
+
+	inputPath := unstructured.StagePath(controllerutils.PipelineNameFromOwner(destinationSyncCR), destinationSyncCR.Spec.Depends[0].Name)
+	filePaths, err := fs.ListFilesInPath(ctx, inputPath)
+	if err != nil {
+		logger.Error(err, "failed to list files in path")
+		return r.handleError(ctx, destinationSyncCR, err)
+	}
+	filterEmbeddingsFiles := unstructured.FilterVectorEmbeddingsFilePaths(filePaths)
```
PuneetPunamiya: Why are we just filtering embedding files ?

api/v1alpha1/unstructureddatapipeline_types.go
concaf: Done — renamed to `dependsOn` with `StageDependency{Name, Type}` struct. Each dependency now carries both the stage name and its type.

api/v1alpha1/unstructureddatapipeline_types.go
concaf: Done — flattened `SourceCrawlerStageConfig.SourceConfig` into `SourceCrawlerConfig` directly. No more nesting.

api/v1alpha1/unstructureddatapipeline_types.go
concaf: Done — `SourceCrawlerStageConfig` struct has been removed. Config fields are directly on `SourceCrawlerConfig`.

api/v1alpha1/unstructureddatapipeline_types.go
concaf: Done — `type` is now directly on `SourceCrawlerConfig`, no `sourceConfig` wrapper.

api/v1alpha1/unstructureddatapipeline_types.go
concaf: Done — moved to a unified secret model: one `secretRef` on `ControllerConfig` (operator-level: filestore AWS, docling key, embedding creds) and one `secretRef` on `UnstructuredDataPipeline` (pipeline-level: source/destination AWS creds). No per-stage secrets.

api/v1alpha1/unstructureddatapipeline_types.go
concaf: Done — `S3Config` no longer has its own `SecretRef`. Credentials come from the pipeline-level unified secret.

api/v1alpha1/chunksgenerator_types.go
concaf: Done — removed `pipelineName`/`DataPipeline` field from all stage CRD specs. Pipeline name is now derived from the controller owner reference via `ParentPipelineNameFromOwnerReference()`.

api/v1alpha1/documentprocessor_types.go
concaf: Done — `DataPipeline` field removed from DocumentProcessorSpec.

api/v1alpha1/vectorembeddingsgenerator_types.go
concaf: Done — `DataPipeline` field removed from VectorEmbeddingsGeneratorSpec.

api/v1alpha1/sourcecrawler_types.go
concaf: Done — `DataPipeline` field removed from SourceCrawlerSpec.

api/v1alpha1/unstructureddatapipeline_types.go
concaf: Not pursuing in this PR — the DAG model intentionally allows arbitrary stage wiring (e.g. SourceCrawler → VEG directly). Hardcoding type-specific dependency rules would break the BYO pipeline flexibility. We can revisit this as an opt-in validation later if needed.

internal/controller/unstructureddatapipeline_controller.go
concaf: Done — simplified using `ListStages()` registry + `meta.EachListItem` for generic iteration. Went from 80+ lines of per-type boilerplate to a clean 4-step algorithm.

internal/controller/documentprocessor_controller.go
concaf: Done — added doc comment to `SetupWithManager` explaining the watch pattern: spec changes trigger via `GenerationChangedPredicate`, watches on other stage types trigger reconcile when an upstream dependency's status changes.

pkg/unstructured/file.go
concaf: No change needed — `StagePath()` takes `pipelineName` as a function parameter (derived at runtime from the owner reference). The function signature didn't change, just where the value comes from.

internal/controller/sourcecrawler_controller.go
concaf: Done — added `// +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch` to both SourceCrawler and DestinationSyncer controllers. Regenerated RBAC role.

internal/controller/sourcecrawler_controller.go
concaf: Acknowledged — this is by design for the flexible DAG model. SourceCrawler is typically a root node with no `dependsOn`, so the watches on downstream types are no-ops (`findDependents` returns empty). Keeping them allows future use cases where a crawler depends on another stage. The runtime cost is negligible (one List + filter per event, returns immediately with no matches).

internal/controller/sourcecrawler_controller.go
concaf: Done — removed the redundant `logger.Error` calls before `handleError`. The error context is now wrapped with `fmt.Errorf` and logged once inside `handleError`.

internal/controller/sourcecrawler_controller.go
concaf: N/A — the `CheckDependenciesExist` gate was removed entirely. Pipeline controller creates all stage CRs in one reconcile loop with owner references, so by the time a stage controller's Reconcile fires, upstream CRs already exist. The watch mechanism handles re-triggering if needed.

internal/controller/destinationsyncer_controller.go
concaf: Done — DestinationSyncer now syncs outputs from ALL stages listed in `dependsOn`, not just embeddings. Each upstream stage's files go to `prefix/stages/<stageName>/` in the destination bucket.

internal/controller/unstructureddatapipeline_controller.go
```diff
@@ -49,267 +48,300 @@ var (
...
 	client.Client
-	Scheme    *runtime.Scheme
-	fileStore *filestore.FileStore
+	Scheme *runtime.Scheme
 }
 
 // +kubebuilder:rbac:groups=operator.dataverse.redhat.com,namespace=unstructured-controller-namespace,resources=unstructureddatapipelines,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=operator.dataverse.redhat.com,namespace=unstructured-controller-namespace,resources=unstructureddatapipelines/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=operator.dataverse.redhat.com,namespace=unstructured-controller-namespace,resources=unstructureddatapipelines/finalizers,verbs=update
+// +kubebuilder:rbac:groups=operator.dataverse.redhat.com,namespace=unstructured-controller-namespace,resources=sourcecrawlers,verbs=get;list;watch;create;update;patch
```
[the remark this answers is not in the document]
concaf: Done — added `delete` verb to RBAC markers for all 5 stage resources. Regenerated role.

internal/controller/sourcecrawler_controller.go
```diff
@@ -0,0 +1,259 @@
...
+		}
+	default:
+		return r.handleError(ctx, sourceCrawlerCR, fmt.Errorf("unsupported source type: %s", sourceCrawlerConfig.Type))
+	}
+
+	storedFiles, err := source.SyncFilesToFilestore(ctx, r.fileStore)
+	if err != nil {
+		logger.Error(err, "failed to store files to filestore")
+		return r.handleError(ctx, sourceCrawlerCR, err)
+	}
```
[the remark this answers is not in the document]
concaf: Done — added `+kubebuilder:rbac` marker for secrets to both SourceCrawler and DestinationSyncer controllers in a previous commit.

internal/controller/unstructureddatapipeline_controller.go
```diff
@@ -49,267 +48,300 @@ var (
...
-
-	// all done, let's update the status to ready
-	successMessage := fmt.Sprintf("successfully reconciled unstructured data product: %s", dataProductName)
-	if err := controllerutils.StatusPatch(ctx, r.Client, unstructuredDataPipelineCR, func() {
-		unstructuredDataPipelineCR.UpdateStatus(successMessage, nil)
-	}); err != nil {
-		logger.Error(err, "failed to update UnstructuredDataPipeline CR status", "namespace", unstructuredDataPipelineCR.Namespace, "name", unstructuredDataPipelineCR.Name)
-		return r.handleError(ctx, unstructuredDataPipelineCR, err)
+	default:
+		return fmt.Errorf("unknown stage type: %s", stage.Type)
```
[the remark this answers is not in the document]
concaf: Done — fixed both issues: (1) nil config check is handled by CEL validation markers on PipelineStage (`XValidation` rules enforce type-config pairing), so the API server rejects mismatches before the controller sees them. (2) Labels are now merged instead of overwritten — `cr.Labels[PipelineLabel] = name` with nil-map init.

internal/controller/destinationsyncer_controller.go
```diff
@@ -0,0 +1,221 @@
...
+		var destination unstructured.Destination
+		switch destConfig.Type {
+		case operatorv1alpha1.TypeS3:
+			destination, err = setupS3DestinationFromConfig(ctx, r.Client, destinationSyncCR.Namespace, destinationSyncCR.Spec.SecretRef, destConfig, dep.Name)
+			if err != nil {
+				return r.handleError(ctx, destinationSyncCR, err)
+			}
+		default:
+			return r.handleError(ctx, destinationSyncCR, fmt.Errorf("unsupported destination type: %s", destConfig.Type))
+		}
```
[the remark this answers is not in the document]
concaf: Done — S3 client and credentials are now created once outside the deps loop. Each iteration only builds a lightweight `S3Destination` struct with the per-dep `StageName`.

pkg/unstructured/source.go
```diff
@@ -200,3 +195,28 @@ func (s *S3BucketSource) storeFile(ctx context.Context, fs *filestore.FileStore,
...
+}
+
+// filestorePath remaps an S3 key to the filestore output directory.
+func (s *S3BucketSource) filestorePath(s3Key string) string {
+	if s.OutputDir == "" {
+		return s3Key
+	}
+	baseName := strings.TrimPrefix(s3Key, s.Prefix)
+	return s.OutputDir + baseName
+}
```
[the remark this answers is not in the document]
concaf: Done — switched `filestorePath` and `s3Key` to use `path.Join` instead of string concatenation.

api/v1alpha1/stages_utils.go
```diff
@@ -0,0 +1,107 @@
...
+import (
+	"fmt"
+	"strings"
+)
+
+// ValidateStages checks that all dependency references exist and the DAG has no cycles.
+func ValidateStages(stages []PipelineStage) error {
+	stageMap := make(map[string]bool, len(stages))
+	for _, stage := range stages {
+		stageMap[stage.Name] = true
```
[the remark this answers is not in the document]
concaf: Already handled — duplicate stage names are enforced at the CRD level via `+listType=map` + `+listMapKey=name` on the `Stages` field. The API server rejects duplicates before `ValidateStages` runs.

api/v1alpha1/stages_utils.go
```diff
@@ -0,0 +1,107 @@
...
+	}
```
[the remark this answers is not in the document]
concaf: Already handled — duplicate stage names are enforced at the CRD level via `+listType=map` + `+listMapKey=name`. See above.

internal/controller/unstructureddatapipeline_controller.go
```diff
@@ -49,267 +48,300 @@ var (
...
-		unstructuredDataPipelineCR.SetWaiting()
-	}); err != nil {
-		logger.Error(err, "failed to update UnstructuredDataPipeline CR status")
-		return ctrl.Result{}, err
+	for _, stage := range stages {
+		if err := r.ensureChildCR(ctx, stage, stages, unstructuredDataPipelineCR); err != nil {
+			logger.Error(err, "failed to ensure child CR", "stage", stage.Name)
+			return r.handleError(ctx, unstructuredDataPipelineCR, err)
+		}
 	}
```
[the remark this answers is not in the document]
concaf: Not pursuing — `markStageCreated` runs once per stage during initial creation only. On subsequent reconciles, `CreateOrUpdate` returns `unchanged` and `markStageCreated` is a no-op status update. The cost is acceptable for the simplicity it provides vs batching.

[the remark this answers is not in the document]
concaf: See above — not pursuing status batching for the same reason.

internal/controller/vectorembeddingsgenerator_controller.go
```diff
@@ -219,19 +191,13 @@ func (r *VectorEmbeddingsGeneratorReconciler) processChunkedFile(ctx context.Con
...
-			AuthFormat: "Bearer",
-			ModelName:  vectorEmbeddingsGeneratorCR.Spec.VectorEmbeddingsGeneratorConfig.ModelName,
-		})
-	default:
-		return false, fmt.Errorf("unsupported embedding model: %s", vectorEmbeddingsGeneratorCR.Spec.VectorEmbeddingsGeneratorConfig.ModelName)
-	}
+	vegConfig := vectorEmbeddingsGeneratorCR.Spec.VectorEmbeddingsGeneratorConfig
+	embeddingClient = embedding.NewHTTPClient(&embedding.HTTPClientConfig{
+		Endpoint:   embeddingEndpoint,
+		APIKey:     embeddingAPIKey,
```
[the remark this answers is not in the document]
concaf: Not pursuing in this PR — the global `embeddingEndpoint`/`embeddingAPIKey` pattern is pre-existing in upstream main. Default `MaxConcurrentReconciles` is 1 and ControllerConfig reconciles infrequently. Fixing this properly (e.g. fetching per-reconcile) is a broader refactor.

internal/controller/sourcecrawler_controller.go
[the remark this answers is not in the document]
concaf: Not pursuing — SQS poll + S3 list ordering is pre-existing behavior in upstream main. Optimizing to check SQS first is a valid improvement but out of scope for this PR.

pkg/awsclienthandler/sqs_client.go
```diff
@@ -58,3 +60,78 @@ func GetSQSClient() (*sqs.Client, error) {
...
+
+// DrainSQSQueue receives messages from the queue and deletes only those whose
+// S3 event matches the given bucket and prefix. Unrelated messages are left in
+// the queue for other consumers.
+// Returns true if any matching messages were found (used as a wake-up signal).
+func DrainSQSQueue(ctx context.Context, sqsClient *sqs.Client, queueURL, bucket, prefix string) (bool, error) {
+	logger := log.FromContext(ctx)
+	hasMessages := false
+
+	for {
```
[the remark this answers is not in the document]
concaf: Not pursuing — the loop terminates when `ReceiveMessage` returns 0 messages (`WaitTimeSeconds: 0` means no long-polling). Unrelated messages are skipped (not deleted), so they reappear on next receive. In practice the loop exits quickly.

internal/controller/destinationsyncer_controller.go
```diff
@@ -0,0 +1,221 @@
...
+	if !isHealthy {
+		logger.Info("ControllerConfig CR is not ready yet, will try again in a bit ...")
+		return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
+	}
+
+	destinationSyncCR := &operatorv1alpha1.DestinationSyncer{}
+	if err := r.Get(ctx, req.NamespacedName, destinationSyncCR); err != nil {
+		logger.Error(err, "failed to get DestinationSyncer CR")
+		return ctrl.Result{}, err
+	}
```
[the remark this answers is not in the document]
concaf: Not pursuing — `r.Get` returning NotFound for the reconciled CR itself is handled by controller-runtime's queue: the deleted object is dequeued and no further reconciles fire. Adding `IgnoreNotFound` is a minor defensive improvement but not a bug.

internal/controller/chunksgenerator_controller.go
```diff
@@ -113,9 +111,13 @@ func (r *ChunksGeneratorReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
 	r.fileStore = fs
 
-	// first fetch the converted files from the filestore for the data product
-	dataProductName := chunksGeneratorCR.Spec.DataProduct
-	filePaths, err := r.fileStore.ListFilesInPath(ctx, dataProductName)
+	pipelineName, err := controllerutils.ParentPipelineNameFromOwnerReference(chunksGeneratorCR)
+	if err != nil {
+		return r.handleError(ctx, chunksGeneratorCR, err)
+	}
+	r.inputPath = unstructured.StagePath(pipelineName, chunksGeneratorCR.Spec.DependsOn[0].Name)
```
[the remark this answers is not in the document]
concaf: Not adding `MinItems=1` on stage CRD specs — SourceCrawler is a root stage with no dependencies. The `DependsOn[0]` access is guarded by the pipeline controller which always populates dependencies from the DAG. Stage CRs are not user-created directly.

internal/controller/documentprocessor_controller.go
```diff
@@ -131,9 +127,14 @@ func (r *DocumentProcessorReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 
-	// now let's process if there are any new raw files that need to be converted
-	dataProductName := documentProcessorCR.Spec.DataProduct
-	filePaths, err := r.fileStore.ListFilesInPath(ctx, dataProductName)
+	// read from upstream stage directory, write to own stage directory
+	pipelineName, err := controllerutils.ParentPipelineNameFromOwnerReference(documentProcessorCR)
+	if err != nil {
+		return r.handleError(ctx, documentProcessorCR, err)
+	}
+	r.inputPath = unstructured.StagePath(pipelineName, documentProcessorCR.Spec.DependsOn[0].Name)
```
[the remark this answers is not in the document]
concaf: Not adding `MinItems=1` on stage CRD specs — SourceCrawler is a root stage with no dependencies. The `DependsOn[0]` access is guarded by the pipeline controller which always populates dependencies from the DAG. Stage CRs are not user-created directly.

internal/controller/vectorembeddingsgenerator_controller.go
```diff
@@ -97,67 +102,34 @@ func (r *VectorEmbeddingsGeneratorReconciler) Reconcile(ctx context.Context, req
...
-	// now that we have the list of files to process, we may remove the force reconcile label as we are ready to accept more events
-	if err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-		latest := &operatorv1alpha1.VectorEmbeddingsGenerator{}
-		if err := r.Get(ctx, req.NamespacedName, latest); err != nil {
-			return err
-		}
-		return controllerutils.RemoveForceReconcileLabel(ctx, r.Client, latest)
-	}); err != nil {
-		logger.Error(err, "failed to remove force reconcile label from VectorEmbeddingsGenerator CR")
+	r.inputPath = unstructured.StagePath(pipelineName, vectorEmbeddingsGeneratorCR.Spec.DependsOn[0].Name)
```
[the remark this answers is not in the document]
concaf: Not adding `MinItems=1` on stage CRD specs — SourceCrawler is a root stage with no dependencies. The `DependsOn[0]` access is guarded by the pipeline controller which always populates dependencies from the DAG. Stage CRs are not user-created directly.

internal/controller/chunksgenerator_controller.go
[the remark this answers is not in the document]
concaf: Not adding `MinItems=1` on stage CRD specs — SourceCrawler is a root stage with no dependencies. The `DependsOn[0]` access is guarded by the pipeline controller which always populates dependencies from the DAG. Stage CRs are not user-created directly.

internal/controller/documentprocessor_controller.go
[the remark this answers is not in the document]
concaf: Not adding `MinItems=1` on stage CRD specs — SourceCrawler is a root stage with no dependencies. The `DependsOn[0]` access is guarded by the pipeline controller which always populates dependencies from the DAG. Stage CRs are not user-created directly.

internal/controller/vectorembeddingsgenerator_controller.go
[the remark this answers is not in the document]
concaf: Not adding `MinItems=1` on stage CRD specs — SourceCrawler is a root stage with no dependencies. The `DependsOn[0]` access is guarded by the pipeline controller which always populates dependencies from the DAG. Stage CRs are not user-created directly.

api/v1alpha1/unstructureddatapipeline_types.go
```diff
@@ -18,8 +18,53 @@ package v1alpha1
...
+//	    - name: crawl
+//	      type: SourceCrawler
+//	      sourceCrawlerConfig:
+//	        type: s3
+//	        s3Config:
+//	          bucket: data-ingestion-bucket
+//	          prefix: documents/
+//	    - name: convert
+//	      type: DocumentProcessor
+//	      dependsOn: [crawl]
```
concaf:
```suggestion
	      dependsOn:
- name: crawl
  type: SourceCrawler
```

internal/controller/vectorembeddingsgenerator_controller.go
```diff
@@ -219,19 +191,13 @@ func (r *VectorEmbeddingsGeneratorReconciler) processChunkedFile(ctx context.Con
...
-			ModelName:  vectorEmbeddingsGeneratorCR.Spec.VectorEmbeddingsGeneratorConfig.ModelName,
-		})
-	default:
-		return false, fmt.Errorf("unsupported embedding model: %s", vectorEmbeddingsGeneratorCR.Spec.VectorEmbeddingsGeneratorConfig.ModelName)
-	}
+	vegConfig := vectorEmbeddingsGeneratorCR.Spec.VectorEmbeddingsGeneratorConfig
+	embeddingClient = embedding.NewHTTPClient(&embedding.HTTPClientConfig{
+		Endpoint:   embeddingEndpoint,
+		APIKey:     embeddingAPIKey,
+		AuthFormat: "Bearer",
```
vinamra28: I guess we should go with the previous model so that we can limit the usage of embedding models

concaf: The old code had a per-model switch with hardcoded model names (`nomic-ai/nomic-embed-text-v1.5`) and per-model endpoint/apiKey lookups from the secret. We simplified this because the embedding endpoint and API key are now in the operator-level ControllerConfig secret — the model name is passed to the embedding service which handles model selection internally. Adding new models no longer requires code changes.

If we want to restrict which models users can specify, we could add a `+kubebuilder:validation:Enum` marker on the `modelName` field. But hardcoding per-model endpoint lookups in the controller is the wrong layer for that. Thoughts?

concaf: Done — added the model validation switch back. New models can be added as cases to the switch.

vinamra28 (approved):

PR #258, opened 2026-06-30
Fix: SQS Client Initialization

gshikhar2021:
- Previously it was returning global variable eventually not calling NewSQSClientFromConfig()


internal/controller/sourcecrawler_controller.go
```diff
@@ -141,17 +141,17 @@ func (r *SourceCrawlerReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	// determine requeue strategy based on SQS configuration
 	sqsQueueURL := sourceCrawlerConfig.S3Config.SQSQueueURL
 	if sqsQueueURL != "" {
```
concaf, line 143: the client should be initialized here

gshikhar2021: Fixed

concaf (approved):

PR #266, opened 2026-07-03
Use global Google client for group membership resolution

gshikhar2021:
Group membership lookups via Cloud Identity API are not
  per-pipeline — they use a shared controller-level service account
  (CONTROLLER_GOOGLE_SERVICE_ACCOUNT_JSON) instead of individual
  pipeline credentials. Falls back to the per-pipeline client when
  the global client is not configured

### DIFF
--- a/config/samples/unstructured-secret.yaml
+++ b/config/samples/unstructured-secret.yaml
@@ -12,6 +12,17 @@ stringData:
   FILE_STORE_AWS_ENDPOINT: http://localhost:4566
   EMBEDDING_ENDPOINT: "http://localhost:11434/v1/embeddings"
   EMBEDDING_API_KEY: ""
+  CONTROLLER_GOOGLE_SERVICE_ACCOUNT_JSON: |
+    {
+      "type": "service_account",
+      "project_id": "your-project-id",
+      "private_key_id": "key-id",
+      "private_key": "-----BEGIN PRIVATE KEY-----\n...\n-----END PRIVATE KEY-----\n",
+      "client_email": "sa@your-project.iam.gserviceaccount.com",
+      "client_id": "123456789",
+      "auth_uri": "https://accounts.google.com/o/oauth2/auth",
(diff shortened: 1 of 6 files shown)


config/samples/unstructured-secret.yaml
```diff
@@ -12,6 +12,17 @@ stringData:
   FILE_STORE_AWS_ENDPOINT: http://localhost:4566
   EMBEDDING_ENDPOINT: "http://localhost:11434/v1/embeddings"
   EMBEDDING_API_KEY: ""
+  CONTROLLER_GOOGLE_SERVICE_ACCOUNT_JSON: |
```
concaf: rename this to something specific because this is being used for only one purpose which is to resolve groups ... maybe `GROUPS_READER_GOOGLE_SERVICE_ACCOUNT_JSON`

pkg/gdrive/permissions.go
```diff
@@ -138,8 +138,11 @@ func (c *Client) GetFilePermissions(
 							}
 						}
 
-						// Fetch from API
-						members, err := c.googleClient.GetGroupMembers(
+						groupClient := c.googleClient
+						if c.globalGoogleClient != nil {
+							groupClient = c.globalGoogleClient
+						}
```
vinamra28: we don't need this.

internal/controller/sourcecrawler_controller.go
```diff
@@ -221,7 +221,7 @@ func (r *SourceCrawlerReconciler) buildGDriveSource(
 		return nil, errors.New("cache client not initialized in ControllerConfig")
 	}
 
-	gdriveClient, err := gdrive.NewClient(googleClient, LDAPClient, CacheClient)
+	gdriveClient, err := gdrive.NewClient(googleClient, GlobalGoogleClient, LDAPClient, CacheClient)
```
vinamra28, line 228: if GlobalGoogleClient is nil then return the error with appropriate message

vinamra28 (approved):

PR #273, opened 2026-07-03
Increase controller concurrency for parallel reconciliation

concaf:
## Summary
- Refactored 5 reconcilers (DocumentProcessor, SourceCrawler, ChunksGenerator, VectorEmbeddingsGenerator, DestinationSyncer) to move per-reconcile state (`fileStore`, `inputPath`, `outputPath`, `doclingConfig`) from struct fields to local variables and function parameters, eliminating data races when multiple goroutines reconcile concurrently
- Set `MaxConcurrentReconciles` to 10 (default) for all controllers except ControllerConfig (stays at 1 since it manages singleton global state)
- Added `--max-concurrent-reconciles` CLI flag to tune concurrency without rebuilding

### Why this is safe
- controller-runtime guarantees the **same object is never reconciled concurrently** regardless of `MaxConcurrentReconciles` — the workqueue deduplicates by namespace/name key
- External API concurrency is already bounded by existing semaphores (Docling, Langchain)
- Status updates use merge-patch (`StatusPatch`), minimizing conflict surface across different objects
- The controller-runtime client is goroutine-safe

### What changed per controller
| Controller | Struct fields removed | `MaxConcurrentReconciles` |
|---|---|---|
| ControllerConfig | none | 1 (unchanged) |
| UnstructuredDataPipeline | none | 10 |
| DocumentProcessor | `doclingConfig`, `fileStore`, `inputPath`, `outputPath` | 10 |
| SourceCrawler | `fileStore` | 10 |
| ChunksGenerator | `fileStore`, `inputPath`, `outputPath` | 10 |
| VectorEmbeddingsGenerator | `fileStore`, `inputPath`, `outputPath` | 10 |
| DestinationSyncer | `fileStore` | 10 |

## Test plan
- [ ] `go build ./...` passes
- [ ] `go test ./...` passes
- [ ] Deploy with default `--max-concurrent-reconciles=10` and verify multiple CRs of the same kind reconcile in parallel
- [ ] Deploy with `--max-concurrent-reconciles=1` and verify behavior matches current production (no regressions)
- [ ] Monitor API server latency under concurrent reconciliation load

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -63,6 +63,7 @@ func main() {
 	var probeAddr string
 	var secureMetrics bool
 	var watchNamespace string
+	var maxConcurrentReconciles int
 	var enableHTTP2 bool
 	var tlsOpts []func(*tls.Config)
 	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address the metrics endpoint binds to. "+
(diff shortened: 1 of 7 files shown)


cmd/main.go
```diff
@@ -63,6 +63,7 @@ func main() {
 	var probeAddr string
 	var secureMetrics bool
 	var watchNamespace string
+	var maxConcurrentReconciles int
```
concaf, line 66: figure out exponential backoff for all services

PR #280, opened 2026-07-06
Store applied config in stage CR status

concaf:
## Summary
- Adds `appliedConfig` field to `DocumentProcessorStatus`, `ChunksGeneratorStatus`, and `VectorEmbeddingsGeneratorStatus`
- Each controller writes the resolved config (with defaults applied) to status on successful reconciliation
- Makes it easy to inspect what config is actually being used, especially when relying on sane defaults

## Test plan
- [x] `make manifests generate` passes
- [x] `make lint` passes (no new Go lint issues)
- [x] `make test` passes
- [x] Verified no tests, samples, or docs need updating (status fields are controller-populated, not user-facing in samples)


vinamra28 (dismissed):

config/crd/bases/operator.dataverse.redhat.com_chunksgenerators.yaml
```diff
@@ -122,6 +122,60 @@ spec:
           status:
             description: ChunksGeneratorStatus defines the observed state of ChunksGenerator
             properties:
+              appliedConfig:
```
PuneetPunamiya, line 125: We are also planning to add inaccessible items in the status as well in #294, let's say the pipeline has 10 stages the status would be very large and also we need to think of the size of status and I think we should revisit once if we should add the whole config or some minimal thing ?

PR #281, opened 2026-07-06
Add SSO_DISABLE_INTROSPECTION to skip token introspection

PuneetPunamiya:
Disabling introspection lets auth pass bearer tokens through without a cross-environment validation failure


pkg/auth/oauth.go
```diff
@@ -31,27 +31,34 @@ import (
...
+		CallbackURL:          os.Getenv("SSO_CALLBACK_URL"),
+		DisableIntrospection: boolFromEnv("SSO_DISABLE_INTROSPECTION"),
 	}
 	return cfg, nil
 }
 
+// boolFromEnv returns true when the environment variable is set to "true".
+func boolFromEnv(key string) bool {
+	return os.Getenv(key) == "true"
+}
```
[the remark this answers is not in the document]
vinamra28: valid suggestion from gemini

pkg/auth/oauth.go
```diff
@@ -179,6 +188,10 @@ func (m *Middleware) Authenticate(next http.Handler) http.Handler {
 
 // validateToken checks the cache first, then delegates to the Provider.
 func (m *Middleware) validateToken(ctx context.Context, token string) (*IntrospectionResponse, error) {
+	if m.disableIntrospection {
+		return &IntrospectionResponse{Active: true}, nil
+	}
```
vinamra28, line 201: tools calls use this introspection token to fetch the username, can we parse the token and extract the username and send it back into the response?

vinamra28, line 201: quick question, can we also try adding a check that this introspection should be skipped if the issuer is certain host?

cc @concaf

vinamra28 (dismissed):

vinamra28 (approved):

PR #286, opened 2026-07-07
Add get_processed_document MCP tool and chain it from get_chunks

PuneetPunamiya:
## Summary

- New `get_processed_document` MCP tool retrieves full markdown content from the DocumentProcessor stage table by `file_id`
- `get_chunks_for_embeddings` now returns `file_id` in results and prompts the LLM agent to call `get_processed_document` as a follow-up for the full document
- Generalize `GetPipelineQueryConfig` to accept a stage type parameter so it works for both VectorEmbeddingsGenerator and DocumentProcessor stages
- Add `snowflake.GetProcessedDocument()` to query the processed document table
- Extract shared `errOAuthTokenNotFound` constant across MCP tools

## Test plan

- [ ] Verify `get_chunks_for_embeddings` returns `file_id` in chunk results
- [ ] Verify `get_chunks_for_embeddings` response includes the NEXT STEP instruction to call `get_processed_document`
- [ ] Verify `get_processed_document` retrieves the correct document by `file_id` from the DocumentProcessor stage table
- [ ] Verify `GetPipelineQueryConfig` correctly resolves query config for both `StageTypeVectorEmbeddingsGenerator` and `StageTypeDocumentProcessor`
- [ ] Verify error handling for missing pipeline, missing query config, and missing document

### DIFF
--- a/internal/mcp/tools/get_chunks.go
+++ b/internal/mcp/tools/get_chunks.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/modelcontextprotocol/go-sdk/mcp"
+	operatorv1alpha1 "github.com/redhat-data-and-ai/unstructured-data-controller/api/v1alpha1"
 	"github.com/redhat-data-and-ai/unstructured-data-controller/pkg/auth"
 	"github.com/redhat-data-and-ai/unstructured-data-controller/pkg/embedding"
 	"github.com/redhat-data-and-ai/unstructured-data-controller/pkg/k8sclient"
@@ -42,6 +43,7 @@ func RegisterGetChunksForEmbeddings(s *mcp.Server, k8sClient *k8sclient.Client, 
 		Name: "get_chunks_for_embeddings",
 		Description: `Search for relevant text chunks in a pipeline's data product using vector cosine similarity. Returns top 5 matching chunks for the given query.
 If pipeline_name is not known, call list_unstructured_data_pipelines_for_user first and follow the instructions in its response.
+If the returned chunks are not sufficient to answer the user's question, you may call get_processed_document with the same pipeline_name and the file_id from the top matching chunk to retrieve the full processed document for more context.
 On error: report the exact error to the user and STOP. Do NOT retry with other pipelines.
 On follow-up: if the user is not satisfied, ask them which pipeline to search. Do NOT automatically try other pipelines.`,
 	}, func(ctx context.Context, _ *mcp.CallToolRequest, args getChunksArgs) (*mcp.CallToolResult, any, error) {
@@ -66,7 +68,7 @@ On follow-up: if the user is not satisfied, ask them which pipeline to search. D
 		if !ok {
 			log.Error("oauth token not found in context")
 			return &mcp.CallToolResult{
-				Content: []mcp.Content{&mcp.TextContent{Text: "Error: oauth token not found in context"}},
+				Content: []mcp.Content{&mcp.TextContent{Text: errOAuthTokenNotFound}},
 				IsError: true,
 			}, nil, nil
 		}
@@ -79,7 +81,7 @@ On follow-up: if the user is not satisfied, ask them which pipeline to search. D
 			}, nil, nil
 		}
 
-		qc, err := k8sClient.GetPipelineQueryConfig(ctx, args.PipelineName)
+		qc, err := k8sClient.GetPipelineQueryConfig(ctx, args.PipelineName, operatorv1alpha1.StageTypeVectorEmbeddingsGenerator)
 		if err != nil {
 			log.Error("failed to get pipeline query config", "error", err)
 			return &mcp.CallToolResult{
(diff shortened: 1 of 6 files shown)


internal/mcp/tools/get_chunks.go
```diff
@@ -140,9 +142,21 @@ On follow-up: if the user is not satisfied, ask them which pipeline to search. D
 		}
 
 		log.Info("completed successfully", "pipeline", args.PipelineName, "chunks_found", len(chunks))
+
+		nextStep := fmt.Sprintf(
+			"NEXT STEP: Call get_processed_document with pipeline_name=%q and file_id from the highest-scoring chunk above. Use the full markdown_content to answer the user's question.",
+			args.PipelineName,
```
vinamra28: do we want to force the agent to call the `get_processed_document`? How about we let the agent decide if it wants to get the markdown content for quality improvement? My worry here is that a big file content returned might blow up agent context window. WDYT?

internal/mcp/tools/get_processed_document.go
```diff
@@ -0,0 +1,129 @@
...
+			return &mcp.CallToolResult{
+				Content: []mcp.Content{&mcp.TextContent{Text: fmt.Sprintf("Error resolving pipeline %q: %v", args.PipelineName, err)}},
+				IsError: true,
+			}, nil, nil
+		}
+
+		if qc.Database == "" || qc.Schema == "" || qc.Table == "" {
+			log.Error("pipeline query config has empty fields", "database", qc.Database, "schema", qc.Schema, "table", qc.Table)
+			return &mcp.CallToolResult{
+				Content: []mcp.Content{&mcp.TextContent{Text: fmt.Sprintf("Error: pipeline %q has incomplete Snowflake query configuration for DocumentProcessor stage", args.PipelineName)}},
```
vinamra28: end user might not know about `DocumentProcessor` stage or pipeline, how about we return a message "unable to retrieve the document, please check again in some time"

internal/mcp/tools/get_chunks.go
PuneetPunamiya: Good catch, have updated the PR

internal/mcp/tools/get_chunks.go
```diff
@@ -140,9 +142,23 @@ On follow-up: if the user is not satisfied, ask them which pipeline to search. D
...
+
+		var nextStep string
+		if len(chunks) > 0 && chunks[0].FileID != "" {
+			nextStep = fmt.Sprintf(
+				"\n\nTIP: To retrieve the full processed document, call get_processed_document with pipeline_name=%q and file_id=%q.",
+				args.PipelineName, chunks[0].FileID,
+			)
+		} else if len(chunks) > 0 {
+			nextStep = fmt.Sprintf(
+				"\n\nTIP: To retrieve the full processed document, call get_processed_document with pipeline_name=%q and the file_id from the highest-scoring chunk above.",
```
vinamra28: I wonder if `fileID` field is going to be empty for any record. We should probably return an error in such scenarios. @concaf WDYT?

internal/mcp/tools/get_chunks.go
```diff
@@ -140,9 +142,23 @@ On follow-up: if the user is not satisfied, ask them which pipeline to search. D
 		}
 
 		log.Info("completed successfully", "pipeline", args.PipelineName, "chunks_found", len(chunks))
+
+		var nextStep string
+		if len(chunks) > 0 && chunks[0].FileID != "" {
+			nextStep = fmt.Sprintf(
+				"\n\nTIP: To retrieve the full processed document, call get_processed_document with pipeline_name=%q and file_id=%q.",
+				args.PipelineName, chunks[0].FileID,
+			)
```
vinamra28: wondering if we should return the first file_id to the LLM 🤔

internal/mcp/tools/get_processed_document.go
```diff
@@ -0,0 +1,129 @@
...
+			return &mcp.CallToolResult{
+				Content: []mcp.Content{&mcp.TextContent{Text: errOAuthTokenNotFound}},
+				IsError: true,
+			}, nil, nil
+		}
+
+		if k8sClient == nil {
+			log.Error("kubernetes client is nil")
+			return &mcp.CallToolResult{
+				Content: []mcp.Content{&mcp.TextContent{Text: "Error: kubernetes client is not initialized"}},
```
vinamra28: should we return the error like "unable to fetch the pipelines"

internal/mcp/tools/get_processed_document.go
```diff
@@ -0,0 +1,129 @@
...
+				IsError: true,
+			}, nil, nil
+		}
+
+		qc, err := k8sClient.GetPipelineQueryConfig(ctx, args.PipelineName, operatorv1alpha1.StageTypeDocumentProcessor)
+		if err != nil {
+			log.Error("failed to get pipeline query config", "error", err)
+			return &mcp.CallToolResult{
+				Content: []mcp.Content{&mcp.TextContent{Text: fmt.Sprintf("Error resolving pipeline %q: %v", args.PipelineName, err)}},
```
vinamra28: same here, can we improve the error message?

internal/mcp/tools/get_processed_document.go
PuneetPunamiya: Updated

internal/mcp/tools/get_processed_document.go
```diff
@@ -0,0 +1,129 @@
...
+				Content: []mcp.Content{&mcp.TextContent{Text: fmt.Sprintf("Error fetching processed document: %v", err)}},
+				IsError: true,
+			}, nil, nil
+		}
+
+		jsonBytes, err := json.Marshal(doc)
+		if err != nil {
+			log.Error("failed to marshal result", "error", err)
+			return &mcp.CallToolResult{
+				Content: []mcp.Content{&mcp.TextContent{Text: fmt.Sprintf("Error marshaling result: %v", err)}},
```
vinamra28: let's not return the raw error in response

internal/mcp/tools/get_processed_document.go
```diff
@@ -0,0 +1,129 @@
...
+		databaseName := strings.ToUpper(strings.ReplaceAll(qc.Database, "-", "_"))
+		schemaName := strings.ToUpper(qc.Schema)
+		tableName := strings.ToUpper(qc.Table)
+
+		log.Info("querying snowflake", "database", databaseName, "schema", schemaName, "table", tableName, "file_id", args.FileID)
+		doc, err := snowflake.GetProcessedDocument(ctx, oauthToken, databaseName, schemaName, tableName, args.FileID)
+		if err != nil {
+			log.Error("failed to get processed document from snowflake", "error", err)
+			return &mcp.CallToolResult{
+				Content: []mcp.Content{&mcp.TextContent{Text: fmt.Sprintf("Error fetching processed document: %v", err)}},
```
vinamra28: same as above.

vinamra28 (approved):

PR #291, opened 2026-07-08
Replace SQS singleton with per-pipeline client cache with TTL eviction

gshikhar2021:


cmd/main.go
concaf, line 1: We should move to a model where, even for S3 and SQS, we should create a new client when the reconciler starts, and immediately after creating, we should add a defer statement that closes the client whenever the function exits or the reconciler stops. We don't need a cache

